PR #41, opened 2015-01-16
Merge comments too

mlocati:


mlocati:
Thank you very much!

mlocati:
Maybe we should also add a `array_values` call around the `array_unique`, just to keep the array with consecutive indexes (for instance see http://codepad.org/V2qkmTwj )

oscarotero:
Yeap, that should be fine

PR #43, opened 2015-01-16
Avoid warning while reading an empty file

mlocati:
When reading an empty file, fread throws this warning:
`Warning: fread(): Length parameter must be greater than 0 in ...Extractors/Extractor.php on line 84`

Let's fix this :wink:


oscarotero:
Great. Thank you!

mlocati:
Thanks!

PR #46, opened 2015-01-19
Fix mo generation

mlocati:
Sorry, I did a little mistake... This fixes it.

### DIFF
--- a/src/Generators/Mo.php
+++ b/src/Generators/Mo.php
@@ -20,11 +20,14 @@ class Mo extends Generator implements GeneratorInterface
         }
         foreach ($translations as $translation) {
             if ($translation->hasContext()) {
-                $haskKey = $translation->getContext()."\x04".$translation->getOriginal();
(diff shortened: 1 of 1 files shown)


mlocati:
:ok:

mlocati:
The problem here is that I'm not sure if a translatable string is identified by context+original or by context+original+plural.
This PR adopts the second approach, that's compatible with what we had before all those PRs of mine :wink:

oscarotero:
I guess the second approach is the right one, but I'm not sure. :+1:

mlocati:
> I guess the second approach is the right one, but I'm not sure. 

Me too. From the [gettext docs](http://www.gnu.org/software/gettext/manual/gettext.html#MO-Files) we have:
> only the singular of the original string takes part in the hash table lookup

We may think from that sentence that we can't have two strings with the same context+original, even if one is singular and the others have plurals... But since we don't use the hash table let's skip this (rare) argument :wink:

PR #47, opened 2015-01-19
Check php 5.6 and hhvm too

mlocati:
Why not? :wink:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: php
 php:
+  - 5.6
   - 5.5
   - 5.4
   - 5.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,5 @@ php:
   - 5.5
   - 5.4
   - 5.3
+  - hhvm
 script: phpunit


oscarotero:
I don't have HHVM installed so I cannot fix any related bug. Let's merge it because currently the library pass the tests, so I guess HHVM does not need special stuff.

mlocati:
> I guess HHVM does not need special stuff.

The code shouldn't be fixed to run with hhvm, because hhvm *should* be a drop-in replacement for php.
This check is useful for two reasons:

1. this gettext library don't hit any of the few incompatibilities between hhvm and php
2. other developers that use hhvm are sure they can safely use this library with their projects

If we meet some hhvm bug, we may update the `.travis.yml` file by adding this lines of code:
```
matrix:
  allow_failures:
    - php: hhvm
```

(almost like [I did for Punic](https://github.com/punic/punic/blob/master/.travis.yml), where [I met a hhvm bug](https://github.com/punic/punic/issues/41) :wink:)

oscarotero:
Ok, I hope that this travis configuration is not needed in the future.

Issue #48, opened 2015-01-20
Message uniqueness


mlocati:
Let's say we have a `test.po` file like this:
```po
msgid ""
msgstr ""
"Project-Id-Version: \n"
"POT-Creation-Date: \n"
"PO-Revision-Date: \n"
"Language: it\n"
"Language-Team: \n"
"Last-Translator: \n"
"Plural-Forms: nplurals=2; plural=(n != 1);\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"

msgid "1 child"
msgstr ""

msgid "1 child"
msgid_plural "2 children"
msgstr[0] ""
msgstr[1] ""
```

If we compile it with `msgfmt` we have:
```
test.po:17: duplicate message definition...
test.po:15: ...this is the location of the first definition
msgfmt: found 1 fatal error
```
So, we may not have in .po files two strings with same context+msgid but different plurals.

I think that we should apply two changes:
1. Revert https://github.com/oscarotero/Gettext/pull/46
2. Add an alternative method to `Translations::find` that does not take in account the plural form

mlocati:
Another problem is with `Translations::mergeWith`. When working with gettext files we should merge entries with the same context+single-form (even if they have different plurals)

oscarotero:
I see. I'm not agree with this behaviour, but the library should be compatible with gettext (as its name says) so this change is necessary, but this affects to different things:

* What happens on scan a po/phpcode/etc source and two entries with the same id but different plural are found? Should it throw an exception? or overwrite it?
* The same behaviour that above should be applied to `Translations::find`
* `Translations::find` should remove the plural argument, or make it optional (only take in account if its defined). But if plurals are not part of the id, I vote for remove it. Maybe an array as third argument with different properties to filter, for example:
```php
//Search only by context + msgid
$t = $translations->find(null, 'comment');

//Search by context + msgid + other non-unique parameters
$t = $translations->find(null, 'comment', array(
    'plural' => 'comments',
    'translation' => 'comentario',
    'plural_translation' => 'comentarios'
));

**Comment by mlocati** [comment]
> What happens on scan a po/phpcode/etc source and two entries with the same id but different plural are found? Should it throw an exception? or overwrite it?

I did a test with these two files (I don't write here the headers to be more clear):
`singular.po`:
```po
msgid "1 child"
msgstr ""
```
`plural.po`:
```po
[...]

PR #49, opened 2015-01-20
Consider messages as unique if they only have same context+single

mlocati:
Plural forms should not be considered when determining message uniqueness in gettext files.

See https://github.com/oscarotero/Gettext/issues/48


oscarotero:
I merged this by now, but there is more work to do before create a new release, according wit #48. I'm a bit bussy right now but I hope to have some time today or tomorrow to do the remaining fixes. Thank you.

mlocati:
Great!

PR #50, opened 2015-01-20
Add methods to create Translations with different plural rules

mlocati:
gettext/gettext currently can extracts strings from source files, thus creating an dictionary of translations (in GNU-gettext terms it may be seen as a POT - Portable Object Template - file).

When creating translations for other languages, we may need to change the plural count of the translations. So, let's introduce this new `setPluralCount` method.

For instance, here's a whole process to create two translation (.po) files, one for Russian (3 plural rules) and one for Japanese (1 plural rule):
```php
$pot = \Gettext\Translations::fromPhpCodeFile($file);
$ruPo = clone $pot;
$ruPo->setPluralCount(3);
$jaPo = clone $pot;
$jaPo->setPluralCount(1);
```

### DIFF
--- a/src/Translation.php
+++ b/src/Translation.php
@@ -364,4 +364,28 @@ class Translation
             $this->flags = array_values(array_unique(array_merge($translation->getFlags(), $this->flags)));
         }
     }
+
+    /**
+     * Changes the plural count of this translation. Please remark that partial plural translations will be emptied.
+     * @param int $plurals
+     */
+    public function setPluralCount($nplurals)
+    {
+        if ($this->hasPlural()) {
+            $newArraySize = $nplurals - 1;
+            if ($newArraySize < 1) {
+                $this->pluralTranslation = array();
+            } else {
+                $oldArraySize = count($this->pluralTranslation);
+                if ($newArraySize < $oldArraySize) {
+                    $this->pluralTranslation = array_slice($this->pluralTranslation, 0, $newArraySize);
+                } elseif ($newArraySize > $oldArraySize) {
+                    $this->translation = '';
+                    $this->pluralTranslation = array_fill(0, $newArraySize, '');
+                }
+            }
+        } else {
+            $this->pluralTranslation = array();
+        }
+    }
 }
--- a/src/Translations.php
+++ b/src/Translations.php
@@ -235,4 +235,15 @@ class Translations extends \ArrayObject
             $this->exchangeArray($filtered);
         }
     }
+
+    /**
+     * Changes the plural count of all the translations. Please remark that partial plural translations will be emptied.
+     * @param int $plurals
+     */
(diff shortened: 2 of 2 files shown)


oscarotero:
The number of plurals is defined on export to po format (for example: https://github.com/oscarotero/Gettext/blob/master/tests/files/po.po#L11) so I'm not sure about this change, because it's a gettext specific rule, and this library tries to be something more abstract that fits more formats. I think a better approach is generate this empty values on export to po, according with the plural form header. For example:
```php
$pot = \Gettext\Translations::fromPhpCodeFile($file);

$pot->setHeader('Plural-Forms', 'nplurals=3; ...');
$pot->toPoFile('locales/ru/locales.po');

$pot->setHeader('Plural-Forms', 'nplurals=1; ...');
$pot->toPoFile('locales/ja/locales.po');
```
This avoid duplications and possible bugs, for example:
```php
$pot = \Gettext\Translations::fromPhpCodeFile($file);

$pot->setHeader('Plural-Forms', 'nplurals=3; ...');
$pot->setPluralCount(2);  //this is wrong because does not match with the plural-forms header
```

mlocati:
IMHO it's quite wrong to fix the number of plural rules just when exporting. What if we'd need to translate Russian or Japaneese from strings that have been extracted with fromPhpCodeFile? We'd need to set the plural rule of the Translations instance, save it, reload it and allow translations to be modified... Not so smart.
Furthermore, we could integrate the plural rules into Gettext. I'm currently developing a helper class to accomplish that (see https://github.com/mlocati/concrete5-translation-library/blob/pot-are-different-from-po/src/Gettext.php ). But it'd be useful for others to have such a function. So, we may have something lime this:
```php
$pot = \Gettext\Translations::fromPhpCodeFile($file);
$po = $pot->createLanguageTranslations('it');
```
That createLanguageTranslations could set the number of plurals as well as adding the `Language` and `Plural-Forms` headers... Really useful!

oscarotero:
Ok, I understand your point. I think the language must be managed from the `Translations` class, affecting not only to each translation but also the header, to avoid inconsistencies. For example:

```php
$po = \Gettext\Translations::fromPhpCodeFile($file);
$po->definePlural(2, 'plural=(n != 1)');

$po->getHeader('Plural-Forms'); // nplurals=2; plural=(n != 1)
```

oscarotero:
Your helper class is great. If you don't mind, I'll include in my library because it's really usefull

mlocati:
:ok: Sure!
We could add two new methods to set plural number/stuff.
One automatic method that accepts a language code and uses the data taken from my library. It returns `false` if the language code is unknown.
The second method is the one that you suggested above...

oscarotero:
Yes, that's what I was thinking. I merge your PR before add these functions.

PR #52, opened 2015-01-21
Allow getLocaleInfo to accept variants

mlocati:
Allow getLocaleInfo to be called with a complex locale identifier.
As for the Unicode specs, locale identifiers may be composed by:
- language code (2 to 3 letters)
- script (4 letters)
- territory (2 letters or 3 numbers)

This PR parses the locale code and extract the language code (mandatory), the script (optional) and the territory (optional).
So, it can accept codes like
- `'it'`
- `'it_IT'`
- `'it_Latn'`
- `'it_Latn_IT'`


mlocati:
I also updated the original source of the rules, so that we can track changes (I'll remove this big array from my repo since you added it to yours).

mlocati:
Thanks!

oscarotero:
:+1:

Issue #56, opened 2015-01-21
Default headers


mlocati:
Currently we set the default headers are only set in the [Po generator](https://github.com/oscarotero/Gettext/blob/c701efc6f88385471ad7d6f34ad11c5301cc01f5/src/Generators/Po.php#L15-L23).
IMHO we should move them to the Translations constructor. This because the default headers should be included in the generated .mo files.
For instance, if we have:
```php
$t = \Gettext\Translations::fromPhpCodeFile($file);
Generators\Mo::toFile($t, $file);
```
The resulting .mo file will be broken since it does not contain any header.

Initializing the default headers in the class constructor will simplify the whole process

PR #57, opened 2015-01-21
Fix #55 and #56

### LINKED ISSUE #55: Headers handling
IMHO we should remove the $context and $language variables from the `Translations` class.
They should always be included only in the $headers variable.
If we do so, we can simplify the extractors.
For instance, we could remove these lines: https://github.com/oscarotero/Gettext/blob/c701efc6f88385471ad7d6f34ad11c5301cc01f5/src/Extractors/Po.php#L41-L48
Furthermore, if we do so, we prevent strange situations where we have a $language set to something and $header['Language'] set to something else.

What do you think?

mlocati:


src/Generators/Po.php
```diff
@@ -12,18 +12,8 @@ public static function toString(Translations $translations)
...
-            'Language-Team' => '',
-            'MIME-Version' => '1.0',
-            'Content-Type' => 'text/plain; charset=UTF-8',
-            'Content-Transfer-Encoding' => '8bit',
-            'Language' => $translations->getLanguage(),
-        ), $translations->getHeaders());
-
-        $headers['POT-Creation-Date'] = $headers['PO-Revision-Date'] = date('c');
+        $headers = $translations->getHeaders();
+        $headers['PO-Revision-Date'] = date('c');
```
mlocati: Since Translations may be a .po and not a .pot file, we shouldn't change the 'POT-Creation-Date' header while we save the file.
'POT-Creation-Date' is initialized to the current date when we initialize the Translations instance, but can be overwritten by the .po/.mo extractors.

tests/PoExtractorTest.php
```diff
@@ -47,8 +47,8 @@ public function testAutomaticHeaders()
         //Extract po with no language/domain headers
         $translations = Gettext\Extractors\Po::fromFile(__DIR__.'/files/plurals.po');
 
-        $this->assertNull($translations->getLanguage(), 'Something erroneously set for language');
-        $this->assertNull($translations->getDomain(), 'Something erroneously set for domain');
+        $this->assertEmpty($translations->getLanguage(), 'Something erroneously set for language');
```
mlocati: Those functions now return an empty string and not null.

oscarotero: I think they should return null if the header is not defined and empty string if its defined but emptied

mlocati: :+1:

src/Translations.php
```diff
@@ -245,18 +281,25 @@ public function mergeWith(Translations $translations, $method = null)
             $method = self::MERGE_ADD | self::MERGE_HEADERS | self::MERGE_COMMENTS;
         }
 
-        if (!$this->getLanguage()) {
```
mlocati: Shall we move these lines under `if ($method & self::MERGE_HEADERS) {` ?

oscarotero: Yes, I think its more appropiate place.

tests/PoExtractorTest.php
mlocati: I left the empty `Language` header in the class initialization because `msginit` creates empty .po files with it.

src/Translations.php
```diff
@@ -163,13 +185,25 @@ public function setLanguage($language)
     }
 
     /**
+     * Checks whether the language is empty or not
+     *
+     * @return boolean
+     */
+    public function hasLanguage()
+    {
+        $language = $this->getDomain();
```
oscarotero: Here should be a `getLanguage()` :)

tests/PoExtractorTest.php
oscarotero: :+1:

src/Translations.php
mlocati: PR updated

mlocati:
Ok, I'm done wil all the suggestions :wink:

mlocati:
I updated this PR:
- MERGE_HEADERS is not set by default
- MERGE_REFERENCES is set by default
- existing translations gets updated only if they are of the same kind (plural with plural, single with single), so that we don't change the type of the existing Translation instances
- we keep the consistency of the number of plurals

mlocati:
I updated this PR:

- MERGE_HEADERS is again set by default
- MERGE_LANGUAGE added (not set by default)

mlocati:
PS: I added a speed improvement. From [some test I did](https://github.com/punic/punic/pull/39#issuecomment-70867558), strlen is quite slower, so let's try to avoid it when possible.

oscarotero:
Ok. Can I merge this? There is some work to do in `Translations::mergeWith` but I can do it.

mlocati:
Sure!

PR #60, opened 2015-01-27
Faster extraction

mlocati:
When adding string to the `Translations` instance, we call [`Translations::offsetSet`](https://github.com/oscarotero/Gettext/blob/v3.0/src/Translations.php#L100-L118) that in turns calls [`Translation:is`](https://github.com/oscarotero/Gettext/blob/v3.0/src/Translations.php#L303-L307) many times.
If we load 5.000 strings, its inner cycle calls `Translation:is` 12.497.500 times (5000\*(5000-1)/2).
What about adding an option to skip this check?

With the code of this pull request, the checks are skipped by default for the .po and .mo importers (but it's configurable by writing `Gettext\Extractors\Mo::$skipIntegrityChecksOnImport`.

### DIFF
--- a/src/Extractors/Blade.php
+++ b/src/Extractors/Blade.php
@@ -13,10 +13,11 @@ class Blade extends Extractor implements ExtractorInterface
     /**
      * {@inheritDoc}
      */
-    public static function fromString($string, Translations $translations = null, $file = '')
+    protected static function fromStringDo($string, Translations $translations, $file)
     {
-        $string = (new BladeCompiler(new Filesystem(), null))->compileString($string);
+        $bladeCompiler = new BladeCompiler(new Filesystem(), null);
+        $string = $bladeCompiler->compileString($string);
 
-        return PhpCode::fromString($string, $translations, $file);
+        PhpCode::fromString($string, $translations, $file);
     }
 }
--- a/src/Extractors/Extractor.php
+++ b/src/Extractors/Extractor.php
@@ -27,7 +27,7 @@ abstract class Extractor
             $translations = new Translations();
         }
         $originalSIC = $translations->getSkipIntegrityChecks();
-        if (self::$skipIntegrityChecksOnImport) {
+        if (static::$skipIntegrityChecksOnImport) {
             $translations->setSkipIntegrityChecks(true);
         }
         try {
@@ -62,7 +62,7 @@ abstract class Extractor
             $translations = new Translations();
         }
         $originalSIC = $translations->getSkipIntegrityChecks();
-        if (self::$skipIntegrityChecksOnImport) {
+        if (static::$skipIntegrityChecksOnImport) {
             $translations->setSkipIntegrityChecks(true);
         }
         try {
--- a/src/Extractors/Extractor.php
+++ b/src/Extractors/Extractor.php
@@ -8,6 +8,12 @@ use Gettext\Translations;
 abstract class Extractor
 {
     /**
+     * Set to true if integrity checks should be skipped diring the import process, false otherwise
+     * @var bool
+     */
+    public static $skipIntegrityChecksOnImport = false;
+
+    /**
      * Extract the translations from a file
      *
      * @param array|string      $file         A path of a file or files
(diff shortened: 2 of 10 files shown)


src/Extractors/Extractor.php
```diff
@@ -20,14 +26,61 @@ public static function fromFile($file, Translations $translations = null)
         if ($translations === null) {
             $translations = new Translations();
         }
+        $originalSIC = $translations->getSkipIntegrityChecks();
+        if (static::$skipIntegrityChecksOnImport) {
```
mlocati: Use `static` instead of `self` to use the class-specific values (see http://3v4l.org/rWAq8 )

src/Extractors/Blade.php
```diff
@@ -13,10 +13,11 @@ class Blade extends Extractor implements ExtractorInterface
     /**
      * {@inheritDoc}
      */
-    public static function fromString($string, Translations $translations = null, $file = '')
+    protected static function fromStringDo($string, Translations $translations, $file)
     {
-        $string = (new BladeCompiler(new Filesystem(), null))->compileString($string);
+        $bladeCompiler = new BladeCompiler(new Filesystem(), null);
+        $string = $bladeCompiler->compileString($string);
```
mlocati: I splitted the constructor in its own line for PHP 5.3 compatibility. See http://3v4l.org/ZANaO vs http://3v4l.org/m9vk9

mlocati:
I tested decoding a .mo file with 4036 entries and the load time dropped from 5.4 seconds to to 0.06 seconds

mlocati:
The load time of the same file but in .po format dropped from 5.7 seconds to 0.24

oscarotero:
I can see the problem but I have a simpler and better solution.
The previous versions of this library stored de translations with numerical keys because duplicated translations was allowed. As now this is not possible because they are merged, we can use context+original to generate unique keys, and avoid the foreach in the find() method. A simple implementation as example:

```php
public function find($context, $original)
{
    $key = Translation::getKey($context, $original);

   return isset($this[$key]) ? $this[$key] : null;
}
```

mlocati:
> [...] avoid the foreach in the find() method. A simple implementation as example:
> ```php
> public function find($context, $original)
> {
>     $key = Translation::getKey($context, $original);
> 
>    return isset($this[$key]) ? $this[$key] : null;
> }
> ```

Will this work with this syntax?
```php
$translations[] = $translation;
```

oscarotero:
Yes, I guess so. `Translations::offsetSet` can modify both key and value before insert. But it has to be checked, obviously.
I'm bussy the next 5 or 6 hours. Fell free to do some tests, if you want, or I'll do it later.

mlocati:
The main problem I can see is if a Translation instance change its $context or $original values.
That way the hash table would contain a wrong key...

oscarotero:
That's a good point. But after do all this work to merge duplicates, manage plurals, etc, I don't like the idea of put all under a setting that by default is false.
One solution should be save the reference of translations in each translation, but I don't like it.
Other solution can be create an "import mode" status. On enter, use the unique string keys on add new elements and on exit convert the string keys to numeric keys.
I have to think about it.

oscarotero:
I was thinking about this and maybe the best solution is that these two values (context and original) was inmutable, asigned only on instantation and without methods setContext() and setOriginal(). If a translation changes the context or the orginal text, it's not longer the same translaton, so this change makes sense. What do you think?

mlocati:
I totally agree :+1:

oscarotero:
Hi, in this comment 35d2db475a38003bc5a686264a03fd807eabfad9 you can see the changes. Now the context and original are inmutable. If you need the same translation with different context and/or original, you can clone it with the method `getClone()`:

```php
$translation = new Gettext\Translation('my-context', 'original-text');
$translation2 = $translation->getClone('new-context');

echo $translation2->getOriginal(); //original-text
```

Can you check it, please? :smiley:

mlocati:
Yes, everything seems to work fine (and fast!)

PS: why did you created v3.1  and then deleted it? PS: in [GitHub](https://github.com/oscarotero/Gettext/releases) v3.1 does not exist anymore, but in [packagist](https://packagist.org/packages/gettext/gettext) it's still there.

oscarotero:
Yes, I accidentally clicked in "publish release" instead "save draft". Today the new version will be finally released

mlocati:
> Today the new version will be finally released

:ok_hand: :+1:

PR #61, opened 2015-01-27
Extract the language and the territory name

mlocati:
What about returning some extra info from Gettext\Utils\Locales::getLocaleInfo()?

With this PR we now have 3 static arrays: one for the plural definition, one for the language names and one for the territory names.


oscarotero:
Great :+1:

PR #62, opened 2015-01-28
Remove development stuff from downloadable archives

mlocati:
With this change, the archives downloaded from https://github.com/oscarotero/Gettext/releases don't contain development stuff


oscarotero:
:+1:

PR #63, opened 2015-01-28
PSR-2 fixes

mlocati:
Just a few spaces to fix


oscarotero:
All right. I always use the php-cs-fixer before each release but thanks anyway :smile:

PR #64, opened 2015-01-28
Add Locales::getLanguages and Locales::getTerritories

mlocati:
We have this data... What about granting access to it? :wink:


mlocati:
:+1:

PR #66, opened 2015-02-09
Fix mo file generation

MarcusDalgren:
Hi,
I just started using this library in one of my applications and everything worked great until I generated the final mo file. It seems that the way it currently works it will generate empty strings for all non translated strings in the mo file. Skipping the untranslated translations seems to fix this and it appears to be what PoEdit is doing. I compared file sizes when using PoEdit and when generating with the Gettext library and they appear to be about the same now with this fix.

### DIFF
--- a/src/Generators/Mo.php
+++ b/src/Generators/Mo.php
@@ -22,6 +22,7 @@ class Mo extends Generator implements GeneratorInterface
         }
 
         foreach ($translations as $translation) {
+            if (!$translation->hasTranslation()) continue;
             if ($translation->hasContext()) {
                 $originalString = $translation->getContext()."\x04".$translation->getOriginal();
             } else {


oscarotero:
Hi, thank for your contribution.
This change seems good for me, but before merge I'd like to know the oppinion of @mlocati because he was who modified the mo generator to use with the native gettext library.

mlocati:
For me it's fine to not include untranslated strings in the .mo file.
BTW, I'd suggest to add an option for that. That's because of the way the Translator component of ZendFramework2 handles the cache: if Translator does not find a translation string, it clears its cache and reloads all the .mo files.
So, if we don't include empty strings in the .mo files, it's like if ZF2 doesn't use the cache.

mlocati:
About ZF2 Translator: here's its logic: https://github.com/zendframework/zf2/blob/release-2.3.4/library/Zend/I18n/Translator/Translator.php#L437-L439

oscarotero:
Ok. Because this is the only generator/extractor with one option, I'll put it as a static value:
```php
Gettext\Generators\Mo::$includeEmptyTranslations = true;
```
When more options like this are required, maybe an $options argument will be added.

mlocati:
:ok_hand:

mlocati:
> Ok. Because this is the only generator/extractor with one option, I'll put it as a static value:
> ```php
> Gettext\Generators\Mo::$includeEmptyTranslations = true;
> ```

What about using it's negated form? Personally I prefer defaults to false...:
```php
Gettext\Generators\Mo::$excludeEmptyTranslations = false;
```

oscarotero:
I think the default behaviour should be don't include the empty translations (as @MarcusDalgren did) because it may be the default behaviour of gettext, (I'm not sure about this). So, the default variable would be `false`:

```php
Gettext\Generators\Mo::$includeEmptyTranslations = false;
```

PR #71, opened 2015-03-20
Added Twig extractor support

exnor:
Note I haven't added tests - not sure how I should handle that given the required libraries are only suggested in composer (means automated testing will fail unless it also pulls down the suggestions).

Happy to add tests if this isn't a problem.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
     },
     "suggest": {
         "illuminate/view": "Is necessary if you want use the Blade extractor",
-        "twig/twig": "Is necessary if you want to use the Twig extractor"
+        "twig/twig": "Is necessary if you want to use the Twig extractor",
+        "twig/extensions": "Is necessary if you want to use the Twig extractor"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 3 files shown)


oscarotero:
Hi, thanks for your great job!
Currently there is another extractor that require third party libraries (Blade extractor) and had no tests. I've made a commit adding tests for this case, using the "require-dev" property to install it only in development contexts. I guess this is the right way to do.

jtreminio:
@oscarotero merging this PR would be great!

exnor:
Aha - no problem then, I will add tests as soon as I can.

exnor:
Added tests :)

oscarotero:
Thank you :+1:

exnor:
Oh, also note (for anyone finding this via Google or whatever) the Twig i18n extension does not appear to support gettext context (hence this not being in the tests above). 

From what I can tell this is something Drupal have added to their framework (but did not add to the Twig source).  I may end up adding it in due course - if so ill update this thread.

jtreminio:
@exnor I actually made Twig work 100% with this project. It's actually a very simple fix, and everything seems to be working. I'll be releasing a Twig extension soon.

exnor:
@jtreminio great!

PR #72, opened 2015-04-28
Avoid adding empty strings

mlocati:
The PhpFunctionsScanner extracts empty strings too (eg `echo __('');`). So, we risk to generate a .po file like this:
```po
# PO headers
msgid ""
msgstr ""
"Header 1: value\n"
"Header 2: value\n"

# Empty string extracted from code
msgid ""
msgstr ""

...
```
This is wrong, since the msgid must be unique (and this specific case messes up the gettext headers).


oscarotero:
Thank you :)

PR #73, opened 2015-04-28
Added functionality to Twig extractor

exnor:
Modified Twig extractor to allow user to add custom Twig extensions should they be required to parse templates for gettext


oscarotero:
Ok, thank you

PR #77, opened 2015-07-11
What is the purpose of the 'glue' character on translation ids

natmchugh:
This is a great library and has saved me loads of time in a project thank-you for creating it and maintaining it. 

The only problem / slight weirdness I have is that I must clean the unprinted glue character 0x04 off the beginning of every msgId before I use it. I have attached a test to show you what I mean.

I guess if you are going to use these hidden characters to indicate something they should be stripped on output but I don't really understand the purpose and removing it didn't fail any tests.

Thanks again.

### DIFF
--- a/tests/PoExtractorTest.php
+++ b/tests/PoExtractorTest.php
@@ -130,4 +130,13 @@ class PoExtractorTest extends PHPUnit_Framework_TestCase
         $t->setPluralTranslation('% estrelas');
         $this->assertTrue($t->hasPluralTranslation());
     }
+
+    public function testKeysAreClean()
+    {
+        $translations = Gettext\Extractors\Po::fromFile(__DIR__.'/files/no-headers.po');
+
+        $msgId = key($translations);
+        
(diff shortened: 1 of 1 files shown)


oscarotero:
The id of each translation is generated using its context and original text [as you can see here](https://github.com/oscarotero/Gettext/blob/master/src/Translation.php#L30). This allows have many translations with the same original text but different context and viceversa. The `\004` character is used to glue both parts. If you see this character in the beginning, it's because there's no context (it's empty).

Why do you want to clean it?

natmchugh:
Ah I see, what I was doing was iterating through my translations and comparing the keys to those from another source.  The files do indeed have no context. I guess what I really wanted was Translation->getOriginal() to get the value to compare.

From my point of view it would make more sense to use a visible character such as "|" to separate the context and msgid as the index since then it would be obvious that it wasn't going to contain the msgid but a compound of context and msgid.

Thanks for the reply.

PR #78, opened 2015-08-13
Update JsonDictionary.php

NoMan2000:
If PHP is set to strict mode, this will give an error sometimes with "Undefined offset of 1".  This checks the value before mapping it, which will remove the false value afterwards.


oscarotero:
Thank you :+1:

PR #79, opened 2015-08-25
Added new js usage that break extractor test and asserts that catch it.

PSF1:
Hi,

I'm using your library into Pharinix to manage translations, but I have problems when extract text from Javascript code. Some times not detect the translation functions, other get a incomplete string.

By this, I changed the JsCodeExtractorTest and the dummy js file to get the same error.

Thanks, I hope will be helpfull.


oscarotero:
Thank you. I'll have a look

oscarotero:
There was a problem with strings containing quotes `"simple ' quote"` and `'double " quote'`. Fixed in 3.5.1. Thanks for the feedback

PSF1:
Thx Oscar, I will update my project : )))))

PR #84, opened 2015-10-28
Update Po.php

Krato:
Function to Escape Control Characters like EOT from strings.

### DIFF
--- a/src/Generators/Po.php
+++ b/src/Generators/Po.php
@@ -49,23 +49,33 @@ class Po extends Generator implements GeneratorInterface
                 $lines[] = 'msgctxt '.self::quote($translation->getContext());
             }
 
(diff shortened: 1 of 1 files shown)


oscarotero:
Thank you :+1:

Krato:
No problem! ;)

mlocati:
I think we have a problem here: the control chars are not escaped, they are simply removed...

mlocati:
@Krato The recent changes to this repository will keep the bytes that you call "Control Characters like EOT"
That's because we can't strip out the bytes in the ranges `\x01-\x09`, `\x0B`, `\x0C`, `\x0E-\x1F`, `x7F` since they may be used in multibyte strings

PR #92, opened 2016-02-16
Test multibyte and tab chars

mlocati:
This will fail...

### DIFF
--- a/src/Generators/Po.php
+++ b/src/Generators/Po.php
@@ -77,7 +77,7 @@ class Po extends Generator implements GeneratorInterface
      */
     private static function quote($string)
     {
-        return '"'.addcslashes($string, "\x00..\x1F").'"';
+        return '"'.addcslashes($string, "\x00..\x1F\"\\").'"';
     }
 
     /**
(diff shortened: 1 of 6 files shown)


mlocati:
...but the next commit should fix this problem :wink:

oscarotero:
Good job. Thank you very much!

mlocati:
I still have to review the po loader, since there too we should be sure to correctly parse the input strings... Tomorrow I will

oscarotero:
:+1:

mlocati:
PS: I just performed the following test on 50+ languages, each with 4000+ strings (including languages that make great use of multibyte chars like Arabic, Greek, Hebrew, Japanese, Persian, Russian, Simplified Chinese, Traditional Chinese, Vietnamese...)

1. Load .po file generated with GNU gettext tools
2. Save to a new .po file
3. compare both files with `msgcat -u` to look for differences between the two

Result? No difference has been found.

That means that `oscarotero/Gettext` correctly reads and writes .po files :wink:

oscarotero:
Amazing :dancers:

PR #93, opened 2016-02-17
Strings conversion refactoring

mlocati:
Let's centralize the string conversions.

I also found another problem when decoding PO strings (see failing test):
```po
msgid "Newline\nSlash and n\\nend"
```

Should be decoded as
```
Newline
Slash and n\nend
```

But is decoded as
```
Newline
Slash and n
end
```


mlocati:
I updated this pull request, fixing the [current `\\` related problem](https://travis-ci.org/oscarotero/Gettext/builds/109804787) and implementing what's described in #95

oscarotero:
Why not put each string function inside the class that use it? I mean:

```php
Extractors\Po::convertString($string);
Extractors\PhpCode::convertString($string);
Generators\Po::convertString($string);

//instead of
String::fromPo($string);
String::fromPhp($string);
String::toPo($string);
```
So we can improve other extractors (like jsCode, for example) without need to modify other helpers classes.

mlocati:
Yes, good idea

mlocati:
Done.

oscarotero:
Great. Thank you. Can this be merged?

mlocati:
Sure!

oscarotero:
:+1:

PR #94, opened 2016-02-17
Disable composer warning about xdebug:

mlocati:
In TravisCI you see this message:

```
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
```

Let's hide it :wink:


oscarotero:
:+1:

PR #99, opened 2016-02-19
Fix extraction of php strings

mlocati:
Fix https://travis-ci.org/oscarotero/Gettext/jobs/110222372

### DIFF
--- a/src/Extractors/PhpCode.php
+++ b/src/Extractors/PhpCode.php
@@ -44,7 +44,7 @@ class PhpCode extends Extractor implements ExtractorInterface
     public static function convertString($value)
     {
         if ($value[0] === "'") {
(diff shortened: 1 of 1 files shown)


mlocati:
1. It's safer to replace the chars after having stripped out the encosing `'`
2. Let's use the PHP 5.3 way to define the arrays :wink:
3. `strtr` is faster than `str_replace`
4. it's better to use `strtr` in this case, see the first answer of http://stackoverflow.com/questions/8177296/when-to-use-strtr-vs-str-replace )

mlocati:
> 4. it's better to use strtr in this case, see the first answer of http://stackoverflow.com/questions/8177296/when-to-use-strtr-vs-str-replace )

A simpler example: https://3v4l.org/O5t0K

oscarotero:
nice!

Issue #100, opened 2016-02-19
Add support for comments in PhpCode extractor


mlocati:
When using xgettext to extract translatable strings from PHP files we have the option to extract comments (the [`-add-comments`](https://www.gnu.org/savannah-checkouts/gnu/gettext/manual/html_node/xgettext-Invocation.html#Operation-mode) option):

For instance, let's say we have this `test.php` file

```php
<?php

echo __(
	/* i18n: %1$s is the partial number, %2$s is the total number. Example: 2 of 3 */
	'%1$s of %2$s',
	$count,
	$total
);

```

If you run this command:
```sh
xgettext --add-comments=i18n --keyword=__:1 --add-location test.php
```

Here's the output:
```po
#. i18n: %1$s is the partial number, %2$s is the total number. Example: 2 of 3
#: test.php:5
#, php-format
msgid "%1$s of %2$s"
msgstr ""
```

What about adding support to extract such comments?

oscarotero:
I'm agree. I knew there were comments in gettext but I didn't know that they had to be placed there.
One question: comments with `//` are allowed too?

mlocati:
From

```php
<?php

echo __(
    /* i18n: Test comment 1 */
	 'test 1'
);

echo __(
    // i18n: Test comment 2
	 'test 2'
);

echo __(
    # i18n: Test comment 3
	 'test 3'
);
```

`xgettext` extracts this

```po
#. i18n: Test comment 1
#: test.php:5
msgid "test 1"
msgstr ""

#. i18n: Test comment 2
#: test.php:10
msgid "test 2"
msgstr ""

#. i18n: Test comment 3
#: test.php:15
msgid "test 3"
msgstr ""
```

oscarotero:
nice!

oscarotero:
I guess this can be closed

mlocati:
> I guess this can be closed

:+1:

PR #102, opened 2016-02-19
Comments extraction from PHP code

mlocati:
Follow-up of #100: what about a code like this?

### DIFF
--- a/src/Utils/PhpFunctionsScanner.php
+++ b/src/Utils/PhpFunctionsScanner.php
@@ -6,7 +6,37 @@ use Gettext\Extractors\PhpCode;
 
 class PhpFunctionsScanner extends FunctionsScanner
 {
+    /**
+     * PHP tokens of the code to be parsed.
+     *
+     * @var array
+     */
     protected $tokens;
+
+    /**
+     * If not false, comments will be extracted.
+     *
+     * @var string|false
+     */
+    protected $extractComments = false;
+
+    /**
+     * Enable extracting comments that start with a tag (if $tag is empty all the comments will be extracted).
+     *
+     * @param string $tag
+     */
+    public function enableCommentsExtraction($tag = '')
+    {
+        $this->extractComments = (string) $tag;
+    }
+
+    /**
+     * Disable comments extraction.
+     */
+    public function disableCommentsExtraction()
+    {
+        $this->extractComments = false;
+    }
 
     /**
      * Constructor.
@@ -15,7 +45,16 @@ class PhpFunctionsScanner extends FunctionsScanner
      */
     public function __construct($code)
     {
-        $this->tokens = token_get_all($code);
+        $this->tokens = array_values(
+            array_filter(
+                token_get_all($code),
+                function ($token)
+                {
+                    return !is_array($token) || $token[0] !== T_WHITESPACE;
+                }
+            )
+        );
+        $this->extractComments = false;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Utils/PhpFunctionsScanner.php
```diff
@@ -6,16 +6,55 @@
...
+    }
+
+    /**
      * Constructor.
      *
      * @param string $code The php code to scan
      */
     public function __construct($code)
     {
-        $this->tokens = token_get_all($code);
```
mlocati: Let's filter out white spaces, so that parsing token is faster

oscarotero:
It looks fine. I think is better to extract always the comments. If the user don't want them, they can be removed later. The rest of the extractors are not configurable. What do you think?

mlocati:
One may want only comments that start with a specific text (like xgettext), otherwise we risk to extract comments that are not related to the strings to be translated.
Think for instance at this example:

```php
// Initialize some variables
$hi = t('Hello');
$x = 1;
$y = 2;
```

So, I think we should keep the same approach as xgettext: only extract comments if requested, and optionally allow to specify a "tag" to filter comments to be extracted.

oscarotero:
ah, so comments can be before the function, not neccesary inside?
Ok, in this case, maybe a static property (like `public static $functions`) ?

mlocati:
Something like https://github.com/mlocati/Gettext/commit/26c3b5302640be451a589e9cff8bc35e8d70b217 ?

oscarotero:
Its fine to me

mlocati:
I forgot to answer this:
> ah, so comments can be before the function, not neccesary inside?

Yes, xgettext can extract comments that are imediately before the function, provided that *"comment blocks supposed to be extracted must be adjacent to keyword lines"* (from [gettext manual](https://www.gnu.org/savannah-checkouts/gnu/gettext/manual/html_node/xgettext-Invocation.html#Operation-mode))

That means that with this code

```php
<?php

echo __(/* i18n: Test comment 1 */'test 1');

/* i18n: Test comment 2 */
echo __('test 2');

/* i18n: Test comment 3 */

echo __('test 3');

/* i18n: Test comment 4 */
echo __(
    'test 4'
);
```

`xgettext` extracts comments 1, 2 and 3, but not the 4th:
```po
#. i18n: Test comment 1
#: test.php:3
msgid "test 1"
msgstr ""

#. i18n: Test comment 2
#: test.php:6
msgid "test 2"
msgstr ""

#. i18n: Test comment 3
#: test.php:10
msgid "test 3"
msgstr ""

#: test.php:14
msgid "test 4"
msgstr ""
```

The current implementation of this pull request is a bit different: it will extracts also `i18n: Test comment 4`, but I don't think this will be a big problem.

mlocati:
Ok, I'll add a few tests and everything is ready to be merged :wink:

mlocati:
Done.

mlocati:
After this has been merged I'll stop boring you and I'll be looking forward to a new release :wink:

oscarotero:
Thanks for your hard work. I'm going to release the new 3.5.9 version soon.

mlocati:
> Thanks for your hard work.

Thanks to you!

> I'm going to release the new 3.5.9 version soon.

Great!

oscarotero:
The wait is over: https://github.com/oscarotero/Gettext/releases/tag/v3.5.9

mlocati:
Super fast! Thank you!!

PR #103, opened 2016-03-10
Register np__() along with the other global functions.

kilahm:
This delegates actual translation to `BaseTranslator::$current->npgettext`

I hope this was just an oversight and there isn't a reason to exclude this particular translation function.

Since this is just an alias, I thought testing would not be helpful.


oscarotero:
Ops, you're right. Thank you!

PR #104, opened 2016-03-20
Json-Generator: Show all strings with pretty print

sourcerer-mike:
The Json-Generator filtered untranslated string out for the JSON-File.
With such file a developer needs more time to translate things
because he needs to look it up several times.

For a better workflow all translatable strings are included in the JSON-File.
With pretty-print (>= PHP 5.4) the JSON is more human readable.

### DIFF
--- a/src/Generators/JsonDictionary.php
+++ b/src/Generators/JsonDictionary.php
@@ -3,6 +3,7 @@
 namespace Gettext\Generators;
 
 use Gettext\Translations;
+use Symfony\Component\Serializer\Encoder\JsonEncoder;
 
 class JsonDictionary extends Generator implements GeneratorInterface
 {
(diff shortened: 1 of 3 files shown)


mlocati:
What happens on PHP 5.3 (`JSON_PRETTY_PRINT` is not defined)?

sourcerer-mike:
Wow. You like to keep that PHP 5.3 support?
PHP itself does not support that anymore. 

I might be able to fix this.

mlocati:
Yes, this library still supports 5.3 (as you can read from the composer.json file)

sourcerer-mike:
Thanks. I contributed to some projects and some like to drop old PHP.
For yours I've taken a workaround from http://php.net/manual/en/function.json-encode.php#80339 

Feel free to merge or not :)

oscarotero:
I think a better way is to allow to configure how the json can be generated, using a public static config variable. For example:

```php
Gettext\Generators\JsonDictionary::$options = JSON_PRETTY_PRINT;
```

oscarotero:
I close this pull request because I've added the `$options` static property to JsonDictionary generator

sourcerer-mike:
Thank you! I didn't had time for that recently due to so much community driven events.
I'll add that to my project https://github.com/sourcerer-mike/wp-easy-translate after all the trouble ;)

PR #105, opened 2016-03-20
Yaml Generator and Extractor

sourcerer-mike:
Yaml extractor and generator (with unit tests)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
     },
     "require": {
         "php": ">=5.3.0",
-        "gettext/languages": "2.*"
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -19,7 +19,8 @@
     },
     "require": {
         "php": ">=5.3.0",
-        "gettext/languages": "2.*"
+        "gettext/languages": "2.*",
+        "symfony/yaml": "~2"
```
oscarotero: This package should be suggested, not required. Not everybody need it.

PR #106, opened 2016-03-20
CSV Generator and Extractor

sourcerer-mike:
CSV files are a good format for customers and non-developers.
To exchange translations with them a csv generator and extractor have been introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ $string = file_put_contents('locales/en.po', $content);
 
 The available generators are:
 
-* `Gettext\Generators\Csv` - Exports to CSV format
+* `Gettext\Generators\CsvDictionary` - Exports to CSV format
 * `Gettext\Generators\Mo` - Exports to Mo format
 * `Gettext\Generators\Po` - Exports to Po format
 * `Gettext\Generators\PhpArray` - Exports to php code that returns an array with all values
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ $string = file_put_contents('locales/en.po', $content);
 
 The available generators are:
 
+* `Gettext\Generators\Csv` - Exports to CSV format
 * `Gettext\Generators\Mo` - Exports to Mo format
 * `Gettext\Generators\Po` - Exports to Po format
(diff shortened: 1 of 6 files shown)


sourcerer-mike:
PHP 5.3 go and get your pension! :/

mlocati:
It's not always that easy to move projects to newer PHP versions, in particular the big and widely used ones

sourcerer-mike:
I know and that's okay. Old projects never realy switch to newer packages. Feel free to drop some support sometimes with a new major release or so ;)
Anyway I fixed it for PHP 5.3

Hope you merge it :) I really like to use it in some projects and have a better base for customers. They know Excel so they can help translating now ;)

src/Generators/CsvDictionary.php
```diff
@@ -0,0 +1,41 @@
...
+        //for a simple json translation dictionary, one domain is supported
+        $values = current($array);
+
+        // remove meta / header data
+        if (array_key_exists('', $values)) {
+            unset($values['']);
+        }
+
+        $tmpFile = tempnam(sys_get_temp_dir(), 'gettext_');
+        $handle = fopen($tmpFile, 'w');
```
oscarotero: Why not use ` fopen('php://temp', 'w') `? Seems more optimized than creating a real file in disk.

README.md
```diff
@@ -196,6 +196,7 @@ $string = file_put_contents('locales/en.po', $content);
 
 The available generators are:
 
+* `Gettext\Generators\Csv` - Exports to CSV format
```
oscarotero: The generator and extractor is called CsvDictionary instead Csv :)

src/Generators/CsvDictionary.php
sourcerer-mike: I don't know if others use that stream already so a file is a fail-safe method.
If `php://temp` will create a new file or stream with every fopen then this is fail-safe as well.

Anyway I change it to the "php://temp" now.

sourcerer-mike:
I read the comments and changed the code. Hope to see an merge soon :)
Thanks for the notes!

oscarotero:
Thanks for your contribution :+1:

PR #110, opened 2016-05-24
missing Prints function result

9neo:


oscarotero:
Hello.
Thank you for the contribution but I've decided to remove these functions in the new 4.0 version due they add an unnecesary overloading, since you can do just `<?= __('string') ?>`.
Note that this is a mayor version (not fully stable yet), with some breaking changes. If you want to keep using these function, use the 3.x versions instead.

9neo:
ok i

PR #127, opened 2016-09-06
New feature: Method/function gettext_noop

hmpf:
While PHP's gettext-library lacks `gettext_noop()`, several other languages (and frameworks) has one (C, C++, and GCC, Lisp, Scheme, perl as per https://www.gnu.org/savannah-checkouts/gnu/gettext/manual/html_node/xgettext-Invocation.html). 

`gettext_noop()` marks a string as translatable, in order to make it discoverable by extractors, without actually translating it. It is a workaround for translating the contents of variables: as the variable is set, its contents are also marked with `gettext_noop()`:

```
# earlier in the code, discoverable by parsing extractors...
$message = gettext_noop('This is the message');

# later in the code...
echo gettext($message);
```
While some language-dependent variant of "gettext-noop" seems to be in use everywhere that has it, the actual look (or necessity) of `N__` is up for debate.

### DIFF
--- a/src/Extractors/PhpCode.php
+++ b/src/Extractors/PhpCode.php
@@ -31,6 +31,7 @@ class PhpCode extends Extractor implements ExtractorInterface
             'np__' => 'npgettext',
             'dnpgettext' => 'dnpgettext',
             'dnp__' => 'dnpgettext',
+            'noop' => 'noop',
         ],
     ];
 
--- a/src/Extractors/PhpCode.php
+++ b/src/Extractors/PhpCode.php
@@ -32,6 +32,7 @@ class PhpCode extends Extractor implements ExtractorInterface
             'dnpgettext' => 'dnpgettext',
             'dnp__' => 'dnpgettext',
             'noop' => 'noop',
+            'noop__' => 'noop',
         ],
     ];
 
--- a/src/BaseTranslator.php
+++ b/src/BaseTranslator.php
@@ -10,6 +10,14 @@ abstract class BaseTranslator implements TranslatorInterface
     /**
      * @see TranslatorInterface
      */
+    public function gettext_noop($original)
+    {
+        return $original;
+    }
+
+    /**
+     * @see TranslatorInterface
+     */
     public function register()
     {
         $previous = self::$current;
--- a/src/BaseTranslator.php
+++ b/src/BaseTranslator.php
@@ -10,7 +10,7 @@ abstract class BaseTranslator implements TranslatorInterface
     /**
      * @see TranslatorInterface
      */
-    public function gettext_noop($original)
+    public function noop($original)
     {
         return $original;
     }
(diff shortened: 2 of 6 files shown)


oscarotero:
Hello. Thanks for your contribution.
I don't understand exactly what's the point of `gettext_noop`. I mean, why I would do this:
```php
$message = gettext_noop('This is the message');
echo gettext($message);
```

if I can do this:

```php
$message = gettext('This is the message');
echo $message;
```

Surely, there are use cases in which this makes sense, but I cannot think of any.

And about the function name, `N__()` cannot be used because it conflicts with `n__()`(already used to plurals). I see you've renamed to `noop()`, but I preffer to use a similar name to the rest of the functions (ending by `__`) in order to prevent possible conflicts, so `noop__()` is better for me. In the `Translator` class I preffer to use camelCase names, to be compatible with PSR-2. Maybe `gettextNoop`?

hmpf:
It's a workaround for not having lazy evaluation in the languages that have it IIUC. It's when you don't know which language to translate into yet.

I'm cleaning up old code in anticipation of a major version bump. We need it for situations like this:

```
__('some message: error ' . $someErrorCode);
```
where `$someErrorCode` won't be known until runtime. I'd rather do it like this:

```
# one place:
$errorMessages = array(
   '1' => noop('some message: error 1'),
   '2' => noop('some message: error 2')
   ..
);

# many places, in different files:
__($errorMessages[$someErrorCode]);
```

Fortunately, every usage of concatenation in the string to translate that I've found can be fixed like this.

oscarotero:
Ok, I understand.
I think you don't need to use this to add more entries to the translates, because this library allows to mix entries from different sources. For example:

```php
//Add entries from files:
$translations = Translations::fromPhpCode($templates);
$translations->addFromYamlFile('errorMessages.yaml');
$translations->addFromPoFile('moreMessages.po');

//Add more entries manually:
$translations->insert('errors', 'some error: error 1');
$translations->insert('errors', 'some error: error 2');
$translations->insert('errors', 'some error: error 3');
```

Anyway, due the `gettext_noop` function is currently used in other languages, I'm ok with merge this. But before, I like to change the function names as commented previously. What do you think about this?

```php
$translator->noop('some message: error 1');
noop__('some message: error 1');
```

And we need to add this funcion to the list of scannable functions in the [jsCode](https://github.com/oscarotero/Gettext/blob/master/src/Extractors/JsCode.php#L14) and [phpCode](https://github.com/oscarotero/Gettext/blob/master/src/Extractors/PhpCode.php#L19) extractors.

Thank you.

src/Extractors/PhpCode.php
```diff
@@ -31,6 +31,7 @@ class PhpCode extends Extractor implements ExtractorInterface
             'np__' => 'npgettext',
             'dnpgettext' => 'dnpgettext',
             'dnp__' => 'dnpgettext',
+            'noop' => 'noop',
         ],
```
oscarotero: We have to include both functions: the Translator method `noop` and the global function (`noop__`)

PR #133, opened 2016-09-21
Comment extractor now acepts multiple strings (v4)

delmicio:


oscarotero:
Please, can you add a test? I'm not sure whether it works. You're using `in_array()` but we have to check if the comment begins with some of the provided strings.

src/Utils/PhpFunctionsScanner.php
```diff
@@ -150,21 +150,18 @@ protected function parsePhpComment($value)
...
-                )
-            ) {
-                $result = $value;
+            if ($value !== '') {
+                if ($this->extractComments === '' || strpos($value, $this->extractComments) === 0) {
+                    $result = $value;
+                }
+                elseif (is_array($this->extractComments)) {
+                    foreach ($this->extractComments as $string) {
+                        if (strpos($value, $string) === 0) {
```
delmicio: @oscarotero as you mentions added the check if starts with string

delmicio:
@oscarotero can you please point me where should I add the test?

oscarotero:
You can use one of the assets existing currenly (for example: https://github.com/oscarotero/Gettext/blob/master/tests/assets/phpcode2/input.php) or create a new with this use case and create a new test case like https://github.com/oscarotero/Gettext/blob/master/tests/TranslationTest.php#L11 to extract and test the result.

oscarotero:
I'll create some test for this. Thank you for the contribution.

PR #136, opened 2017-01-07
Count translate

maxhelias:
I added a new function in the collection "Translations" to return the number of words that were translated.
I also added the tests that go with.

### DIFF
--- a/src/Translations.php
+++ b/src/Translations.php
@@ -417,6 +417,21 @@ class Translations extends ArrayObject
     }
 
     /**
+     * Count all elements translated
+     * 
+     * @return integer
+     */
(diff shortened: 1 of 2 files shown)


maxhelias:
Hi! I see that the Twig_Loader_String class is deprecated and he be removed in Twig 2.0 which is the version used in PHP 7.
What should I do for my tests in PHP 7 ?

oscarotero:
Hi. I think the Twig_Loader_String is a different issue that should be fixed in other commit, so I'm mergin this.
Thank you for the contribution 👍

maxhelias:
Alright, with pleasure. You have a nice project 👍

PR #139, opened 2017-02-04
Adding support for named placeholders in functions

### LINKED ISSUE #138: Named placeholders
I am migrating a big project away from PHP's native gettext and I'd like to use this lib. However we are using named placeholders and it doesn't look like it's possible here. Is there any plan to support named placeholders instead of `sprintf`/`vsprintf` tokens?

For example:

`__('Hello %name%', ['%name%' => $name]);`

Instead of:

`__('Hello %s', $name);`

If not, would you accept a PR to disable including `translator_functions.php` in `BaseTranslator::register`? This way we can use our own `__()` methods which uses `strtr` instead of `vsprintf`. I'm thinking a simple bool argument (default true, to keep backward compatibility) would do the trick, and allow me to manually include my own translator functions after calling `register`.

rlanvin:
When the last argument is an array, strtr will be used instead of vsprintf for all the functions `__()`, `n__()`, etc.

This allow using named placeholders, e.g.

`__('Hello %name%', ['%name' => 'world'])`

Existing syntax using vsprintf is preserved, e.g.

`__('Hello %s', 'world')`

Fix #138

Notes:

- I explicitly added `$args` to the function declaration to avoid calling `func_num_args()` if not necessary. It is marginally faster.
- I didn't find existing tests for all the `*gettext` methods/functions so I added a bunch.

### DIFF
--- a/src/translator_functions.php
+++ b/src/translator_functions.php
@@ -19,7 +19,7 @@ function __($original)
 
     $args = array_slice(func_get_args(), 1);
 
-    return vsprintf($text, is_array($args[0]) ? $args[0] : $args);
+    return is_array($args[0]) ? strtr($text, $args[0]) : vsprintf($text, $args);
 }
 
 /**
@@ -53,7 +53,7 @@ function n__($original, $plural, $value)
 
     $args = array_slice(func_get_args(), 3);
 
-    return vsprintf($text, is_array($args[0]) ? $args[0] : $args);
+    return is_array($args[0]) ? strtr($text, $args[0]) : vsprintf($text, $args);
 }
 
 /**
@@ -74,7 +74,7 @@ function p__($context, $original)
 
     $args = array_slice(func_get_args(), 2);
 
-    return vsprintf($text, is_array($args[0]) ? $args[0] : $args);
+    return is_array($args[0]) ? strtr($text, $args[0]) : vsprintf($text, $args);
 }
 
 /**
@@ -95,7 +95,7 @@ function d__($domain, $original)
 
     $args = array_slice(func_get_args(), 2);
 
-    return vsprintf($text, is_array($args[0]) ? $args[0] : $args);
+    return is_array($args[0]) ? strtr($text, $args[0]) : vsprintf($text, $args);
 }
 
 /**
@@ -117,7 +117,7 @@ function dp__($domain, $context, $original)
 
     $args = array_slice(func_get_args(), 3);
 
-    return vsprintf($text, is_array($args[0]) ? $args[0] : $args);
+    return is_array($args[0]) ? strtr($text, $args[0]) : vsprintf($text, $args);
 }
 
 /**
@@ -140,7 +140,7 @@ function np__($context, $original, $plural, $value)
 
     $args = array_slice(func_get_args(), 4);
 
-    return vsprintf($text, is_array($args[0]) ? $args[0] : $args);
+    return is_array($args[0]) ? strtr($text, $args[0]) : vsprintf($text, $args);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/translator_functions.php
```diff
@@ -9,17 +9,19 @@
  *
  * @return string
  */
-function __($original)
+function __($original, $args = null)
 {
     $text = BaseTranslator::$current->gettext($original);
 
-    if (func_num_args() === 1) {
+    if ( ! $args ) {
```
oscarotero: This weak evaluation makes than `__('%s comments', 0)` does not work.

src/translator_functions.php
```diff
@@ -43,17 +45,19 @@ function noop__($original)
  *
  * @return string
  */
-function n__($original, $plural, $value)
+function n__($original, $plural, $value, $args = null)
 {
     $text = BaseTranslator::$current->ngettext($original, $plural, $value);
 
-    if (func_num_args() === 3) {
+    if ( ! $args ) {
```
oscarotero: Change this, please

src/translator_functions.php
```diff
@@ -64,17 +68,19 @@ function n__($original, $plural, $value)
  *
  * @return string
  */
-function p__($context, $original)
+function p__($context, $original, $args = null)
 {
     $text = BaseTranslator::$current->pgettext($context, $original);
 
-    if (func_num_args() === 2) {
+    if ( ! $args ) {
```
oscarotero: Change this too, please

src/translator_functions.php
```diff
@@ -85,17 +91,19 @@ function p__($context, $original)
  *
  * @return string
  */
-function d__($domain, $original)
+function d__($domain, $original, $args = null)
 {
     $text = BaseTranslator::$current->dgettext($domain, $original);
 
-    if (func_num_args() === 2) {
+    if ( ! $args ) {
```
oscarotero: and this

src/translator_functions.php
```diff
@@ -107,17 +115,19 @@ function d__($domain, $original)
  *
  * @return string
  */
-function dp__($domain, $context, $original)
+function dp__($domain, $context, $original, $args = null)
 {
     $text = BaseTranslator::$current->dpgettext($domain, $context, $original);
 
-    if (func_num_args() === 3) {
+    if ( ! $args ) {
```
oscarotero: and this

src/translator_functions.php
```diff
@@ -130,17 +140,19 @@ function dp__($domain, $context, $original)
  *
  * @return string
  */
-function np__($context, $original, $plural, $value)
+function np__($context, $original, $plural, $value, $args = null)
 {
     $text = BaseTranslator::$current->npgettext($context, $original, $plural, $value);
 
-    if (func_num_args() === 4) {
+    if ( ! $args ) {
```
oscarotero: and this

src/translator_functions.php
```diff
@@ -154,15 +166,17 @@ function np__($context, $original, $plural, $value)
  *
  * @return string
  */
-function dnp__($domain, $context, $original, $plural, $value)
+function dnp__($domain, $context, $original, $plural, $value, $args = null)
 {
     $text = BaseTranslator::$current->dnpgettext($domain, $context, $original, $plural, $value);
 
-    if (func_num_args() === 5) {
+    if ( ! $args ) {
```
oscarotero: and this

src/translator_functions.php
mlocati: I agree: the previous code is much more safe: we may have cases when we pass `null` too, and that should be passes to the vsprintf below

rlanvin:
I changed the functions and added tests for "0" and "null".

oscarotero:
Good job. Thank you!

Issue #140, opened 2017-03-04
Problems with embedded functions in JavaScript source codes


icebird93:
When I tried to extract messages from JavaScript source codes, I bumped into a bug I think. I have simple quoted texts, like:

`i18n.__(‘This is a <a href="javascript:location.reload()">test</a>')`

By the "(" character a new _function_ status is added making the buffer empty and breaking the messages. This is because of a missing status case ('simple-quote').

I fixed it by adding `case 'simple-quote':` in JsFunctionsScanner.php line 105. I do not know if it is deliberately this way or just a bug. I thought this might help others.

oscarotero:
Yes, good catch.
Would you mind create a pull request with this fix?
thank you!

PR #141, opened 2017-03-04
Fixed JavaScript source extraction with single quoted functions

icebird93:
Fixed extracting single quoted messages from JavaScript source codes, as described in issue #140


oscarotero:
Thank you!

PR #145, opened 2017-05-09
Make csv options configurable via $options

Zae:
Make csv options configurable via $options, defaults are set to the PHP defaults.

`Translations::fromCsvFile($destination, ['delimiter' => ';']);`

```
$translations->toCsvFile(
    $pathinfo['dirname'] . '/' . $pathinfo['filename'] . '.csv',
    [
        'includeHeaders' => true,
        'delimiter' => ';'
    ]
);
```

### DIFF
--- a/src/Extractors/Csv.php
+++ b/src/Extractors/Csv.php
@@ -12,17 +12,24 @@ class Csv extends Extractor implements ExtractorInterface
 {
     use HeadersExtractorTrait;
 
+    public static $options = [
+        'delimiter' => ",",
+        'enclosure' => '"',
+        'escape' => "\\"
+    ];
+
     /**
      * {@inheritdoc}
      */
     public static function fromString($string, Translations $translations, array $options = [])
     {
+        $options += static::$options;
         $handle = fopen('php://memory', 'w');
 
         fputs($handle, $string);
         rewind($handle);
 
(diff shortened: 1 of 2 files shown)


Zae:
I see the tests fail because of incompatibilty with the parameter count in older versions of PHP, any suggestions to fix this cleanly?

oscarotero:
A way to make it compatible with php 5.5 is using [version_compare](http://php.net/manual/en/function.version-compare.php) and execute the function without the last argument if the php version is lower than 5.5.4

Zae:
Yes, I was thinking about that, but thought that might be weird that some options won't be available and the code might get a little sloppy with version_compare's all over the place.

I'll have a little think about it and update the PR later :)

Zae:
I added a function that wraps the csv functions with backward compatible versions.

oscarotero:
Thank you 👍

oscarotero:
Ok, I've changed a bit your code. The same options have been extended to csvDictionary (extractor and generator) and, althought there's support for the escape_char argument in `fgetcsv` since php 5.3, I think is better to use the same version than `fputcsv`, in order to maintain the same support in csv generation and extraction.

oscarotero:
v4.4.0 released

PR #149, opened 2017-05-20
Unify plural behaviours between PHP gettext and Translator

### LINKED ISSUE #148: Different behavior between gettext and PHP implementation for unknown plural translations
Note: this is a *very* minor issue you might not want to fix, but I wanted to report it since I came across it during my wrapper unit tests.

Issue: if the translation is unknown PHP's built-in gettext function will use English plural rules (fallback language) while `Translator` (translations loaded from PO file) will use the destination language plural rules.

Example: 

> English > French
> French doesn't pluralize zero, while English does. 

`sandwich` is an unknown key but a French translation is loaded
```php
// with GettextTranslator
n__('sandwich', 'sandwiches', 0); // => sandwiches
// with Translator
n__('sandwich', 'sandwiches', 0); // => sandwich
```

ArthurHoaro:
If the translation key is unknown, we use the English plural rules, as it is done in PHP gettext extension

Fixes #148

### DIFF
--- a/src/Translator.php
+++ b/src/Translator.php
@@ -131,8 +131,8 @@ class Translator extends BaseTranslator implements TranslatorInterface
      */
     public function dnpgettext($domain, $context, $original, $plural, $value)
     {
-        $key = $this->getPluralIndex($domain, $value);
         $translation = $this->getTranslation($domain, $context, $original);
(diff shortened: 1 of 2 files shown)


oscarotero:
Thanks! 👍

PR #153, opened 2017-06-29
Clone translation before merging

lxg:
Clone translation before merging to avoid unintended modification of referenced `Translation` objects.


oscarotero:
Thank you 👍

PR #154, opened 2017-07-26
add fix to skip ' with trailing slash

AleksSem:
add fix to skip ' with trailing slash in translation for example {i18n.__('You don\'t have any bookmark.')}


mlocati:
What if we have an even number of `\` before `'`?

AleksSem:
I don`t check that issue with multiple \ but without that fix ill get 

You don'thaveanybookmark.')}</h5> <p>{i18n.__('Youcanadda .....................

instead {i18n.__('You don't have any bookmark.')}

AleksSem:
Yes you definitely right, your solution i better.

AleksSem:
Should work now

oscarotero:
Thanks 👍

PR #155, opened 2017-08-09
Add dn__ function to both extractors and functions scanner

dorxy:
While extracting translations I noticed the `dn__` function was not working. 
I added it to the applicable extractors and functions scanner switch case.


oscarotero:
Thank you! 👍

PR #156, opened 2017-08-24
php injection - validate eval input from plural forms

soukicz:
There wasn't any input validation on eval'd code in plural forms. Untrusted translation files could create php injection.


oscarotero:
Thank you 👍

PR #157, opened 2017-08-25
fix plural forms validation

soukicz:
My previous PR https://github.com/oscarotero/Gettext/pull/156 missed some characters in plural forms validation

### DIFF
--- a/src/Translations.php
+++ b/src/Translations.php
@@ -234,6 +234,9 @@ class Translations extends ArrayObject
      */
     public function setPluralForms($count, $rule)
     {
+        if (preg_match('/[a-z]/i', str_replace('n', '', $rule))) {
+            throw new \InvalidArgumentException('Invalid Plural form: ' . $rule);
+        }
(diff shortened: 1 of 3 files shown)


oscarotero:
Maybe an easier solution is just check whether the plural form has any letter different than `n`?
Somethin like this (not tested):
```php
if (preg_match('/[A-Ma-mO-Zo-z]/', str_replace('return ', '', $code)) {
    throw new \InvalidArgumentException('Invalid Plural form: ' . $code);
}
```
Just an idea, what do you think?

soukicz:
It would be also better to check it in setter before variable substitution. I will updated it.

oscarotero:
We can also test the regexp with all possible formulas to be sure that doesn't fail with any valid plural formula.

```php
$languages = Gettext\Languages\Language::getAll();

foreach ($languages as $language) {
    //test $language->formula
}
```

soukicz:
I have simplified validation and added testing for predefined plural forms.

oscarotero:
Great. Thank you 👍

PR #159, opened 2017-10-31
don't check Po file content on bigendian

remicollet:
As all Po reference files are LittleEndian

### DIFF
--- a/tests/AbstractTest.php
+++ b/tests/AbstractTest.php
@@ -51,7 +51,10 @@ abstract class AbstractTest extends PHPUnit_Framework_TestCase
         $method = "to{$format}String";
         $content = file_get_contents(static::asset($file.'.'.static::$ext[$format]));
 
-        $this->assertSame($content, $translations->$method(), $file);
+        // Po reference files are LittleEndian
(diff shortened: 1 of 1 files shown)


bowlofeggs:
+1

oscarotero:
Could this fixed using a function like this? http://shiplu.mokadd.im/95/convert-little-endian-to-big-endian-in-php-or-vice-versa/
If so, there's no need to skip tests.

remicollet:
The problem is about reference files provided in the tests/assert directory.

If you don't want to skip this tests, you have to provide both versions (little and big endian)

oscarotero:
Sorry, I don't understand. I've no experience with little/big endian.
What I mean is to convert the string from little-endian to big-endian before the assertion, using the function linked above. If this does not solve the problem, I can merge it anyway, but I'd preffer not to skip tests if it's possible.

remicollet:
```
$ file tests/assets/po/Mo.mo
tests/assets/po/Mo.mo: GNU message catalog (little endian), revision 0.0, 4 messages
```

So you cannot compare file generated by the library (which will be Big Endian), and reference file which is Little Endian.

bowlofeggs:
Thanks!

PR #163, opened 2018-01-23
Ensure global translation functions don't exist before declaring them.

eschalks:
translation_functions.php throws an error when any of the functions it tries to declare already exists. This can cause problems when another Composer package is also trying to declare global functions with the same name.

For the most part this can be worked around by requiring the translation_functions.php file before Composer's autoload.php. However, in some cases this isn't possible. For example, when executing a test case with PHPUnit, because PHPUnit includes Composer's autoload code before the custom bootstrap code.

To fix this I've wrapped all function declarations in an if(function_exists()) statement.

### DIFF
--- a/src/translator_functions.php
+++ b/src/translator_functions.php
@@ -2,190 +2,209 @@
 
 use Gettext\BaseTranslator;
 
-/**
- * Returns the translation of a string.
- *
- * @param string $original
- *
- * @return string
- */
-function __($original)
-{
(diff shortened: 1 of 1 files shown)


oscarotero:
Not sure about this.
I think in case of conflict, there should be an error, not silence it. You can do this check before:

```php
if (!function_exists('__')) {
    Gettext\BaseTranslator::includeFunctions();
}
```

Issue #161, opened 2017-12-14
Improve PHP comments extraction


swissspidy:
Noticed this while working on https://github.com/swissspidy/wp-cli-makepot/issues/4.

Given the following `example.php` PHP file:

```
/* translators: TRANSLATORS!! */
_e( 'hello world', 'test-plugin' );

/* translators: TRANSLATORS!! */
$foo = __( 'foo', 'test-plugin' );
```

The library does not extract the second translator comment.

However, `xgettext --add-comments=translators --keyword=__:1 --add-location example.php` _does_ extract that comment.

It seems like this is a case #102 / #100 didn't test against.

oscarotero:
Ok, thanks for reporting. This requires a deep change in how the comments are extracted here: https://github.com/oscarotero/Gettext/blob/master/src/Utils/PhpFunctionsScanner.php

Currently, only extract comments placed just before the function. In the second case, there's a variable definition between the comment and the function.

I'll try to fix this when I have free time. Anyway, if you want to get involved with a pull request, do not hesitate.

ocean90:
I'm wondering if we can borrow some code from the PHPCS i18n sniff for WordPress even though it depends on the PHPCS' tokenizer: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/blob/735b899c3f2971958656985098022d5c44c52811/WordPress/Sniffs/WP/I18nSniff.php#L632-L729

@jrfnl, maybe you have some thoughts on this?

swissspidy:
I've tried to dig into this in #164. Feedback welcome!

jrfnl:
@ocean90 Thanks for the ping.

For the sniff in WPCS, based on feedback received from various sources, including the maintainer of POEdit (see discussion threads) and tests run with various tools, we ended up with a strict check for a translators comment on the line _**directly above the line containing the gettext function call**_.

The line containing the gettext function call, however, could contain any other token before the gettext call.

So the below examples would all be fine:
```php
/* translators: comment */
echo __( 'string' );

/* translators: comment */
$var = sprintf( __( 'string %d' ), 1 );

printf(
    /* translators: comment */
    __( 'string %d' ),
    1'
);
```

But these wouldn't be:
```php
/* translators: comment not on the line above the gettext call */ __( 'string' );

/* translators: comment not on the line directly above the gettext call */


echo __( 'string' );

/* translators: comment not on the line above the gettext call */
printf(
    __( 'string %d' ),
    1'
);
```



[...]

PR #164, opened 2018-02-20
Parse PHP comments on the line before

swissspidy:
This aims to address #161 by simply adding all comments on the line before to the current function.

It seems to work fine, except that the Twig parser creates files like this:

```php
/*  */
class __TwigTemplate_79a7d3b32ac8bc226e4f26ef9125e6a65dde2b398ad080a22197a3f33e8538b5 extends Twig_Template
{
    public function __construct(Twig_Environment $env)
    {
        parent::__construct($env);

        $this->parent = false;

        $this->blocks = array(
        );
    }

    protected function doDisplay(array $context, array $blocks = array())
    {
        // line 1
        echo "<header>
    <h1>";
        // line 2
        echo gettext("text 1");
        echo "</h1>
</header"...
```

For some reason, these line comments are now being picked up as translator comments and let the tests fail.


swissspidy:
> Now, all these comments are extracted, so I suggest to provide a way to configure whether the phpcode parser should extract the comments or not, in order to disable it on handle php returned by template compilers like twig or blade.

I think it might be even simpler than that.

In `tests/bootstrap.php` there's this line:

```php
Gettext\Extractors\PhpCode::$options['extractComments'] = '';
```

This tells all extractors to extract comments by default, no matter what prefix they have. However, according to http://twig-extensions.readthedocs.io/en/latest/i18n.html, Twig templates will always have the `notes:` prefix in the translator comments, no matter what.

Thus, I'd suggest not to use `$options['extractComments'] = ''` for the Twig extractor and instead use `$options['extractComments'] = 'notes:'`.

See c91726a346a704ca83046ec50abf82d734050556 for an example. Instead of putting that into `bootstrap.php` I'd probably enforce that in `\Gettext\Extractors\Twig::fromString()`.

oscarotero:
Ok, that's great.
Yeah, I think putting this option as default here https://github.com/oscarotero/Gettext/blob/master/src/Extractors/Twig.php#L17 should be enought.

swissspidy:
Added now, although that shouldn't really make any difference :-)

oscarotero:
Thank you very much!

PR #165, opened 2018-02-22
Translator comments should be case insensitive

swissspidy:
I've noticed that in the code base I'm working with the translator comments sometimes start with `translators:` and sometimes with `Translators:`. Instead of passing all possible case variations to the functions scanner, I think it's easier if the extraction happens case insensitive.

This differs from the behavior in `xgettext` but I think makes for a better user experience.


swissspidy:
Closing in favor of #166.

PR #168, opened 2018-02-22
Extend .gitattributes file

swissspidy:
There's no need for `CHANGELOG.md`, `CONTRIBUTING.md` and `phpcs.xml` to be part of the package when installing the library via Composer.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,9 @@
-/tests/      export-ignore
-/phpunit.xml export-ignore
-.*           export-ignore
+/tests/         export-ignore
+/phpunit.xml    export-ignore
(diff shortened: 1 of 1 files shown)


oscarotero:
I disagree.
`CHANGELOG.md` should be provided to any developer and consumer of any sofware. [see this](http://keepachangelog.com/en/1.0.0/)
`CONTRIBUTING.md` is not so important but is nice to have.
`phpcs.xml` can be ignored.

swissspidy:
@oscarotero Fair enough. I now only added `phpcs.xml` to the list.

oscarotero:
Thanks!

PR #170, opened 2018-03-01
Fixed incorrect function description

Amadox:
The function description for dn__ seems to have been copy/pasted from dp__ and wasn't reflecting it's actual purpose.


oscarotero:
Thanks!

PR #175, opened 2018-03-27
Improve Translations return type hinting

ArthurHoaro:
Without return types in `@method` instructions, I have to manually add `@var Translations $foo` to have PHPStorm autocompletion ([also](https://github.com/shaarli/Shaarli/pull/1093#discussion_r175593885) :D).

(EDIT: also the tests are failing since your last merge)


oscarotero:
thanks!

PR #177, opened 2018-04-24
Test enhancement

peter279k:
# Changed log

- Set  the higher PHPUnit version for the ```php-7.2```.
- Use the ```expectedException``` annotation to compatible the different PHPUnit version.
- Let ```testPluralFromValidation``` test method have the ```assertInstanceOf``` assertion and avoid the risky test that don't perform any assertions.


oscarotero:
Thanks!

PR #181, opened 2018-05-15
Specific domain string extraction excluding default domain

briedis:


oscarotero:
Squash merged. Thanks!

PR #184, opened 2018-05-29
VueJS extractor. Include "v-on:" attribute when extracting messages.

briedis:
"v-on:.." attribute was skipped previously, but it can contain translations, this fixes it.


oscarotero:
A better option could be set an option defining the attribute prefixes used. For example:

```php
$prefixes = [':', 'v-on:', 'v-bind:'];
```

This allows to extend or customize the extractor without modify the class.

briedis:
@oscarotero Refactored

oscarotero:
thanks!

PR #188, opened 2018-08-27
VueJS DOM parsing fix

briedis:
When using `$domScript->textContent`, DOM library starts parsing some tags within strings and messes the output up. For example `alert("<span>text</span>");` becomes `alert("<span>text");`
So instead of DOM operations, we extract script contents with a regex.


oscarotero:
Fair enought

oscarotero:
new version available

PR #191, opened 2018-09-09
Added VueJS extractor method docblocks

axlon:
This is just a tiny pull request allowing IDEs like PHPStorm to autocomplete VueJS extractor functions on the Translations class


oscarotero:
👍

PR #193, opened 2018-10-01
Improving byteorder detection on 64 bit environment

DarkNami:
I am not sure this is a PHP bug or not, but PHP cast **2500072158** to **float** on my server. 
(PHP Version 7.2.10-0ubuntu0.18.04.1)
This bug resulting in throw 'Not MO file' exception.
I take WordPress's **get_byteorder()** function as reference, fixed this bug.

windows64
linux64

### DIFF
--- a/src/Extractors/Mo.php
+++ b/src/Extractors/Mo.php
@@ -11,10 +11,6 @@ use Gettext\Utils\StringReader;
  */
 class Mo extends Extractor implements ExtractorInterface
 {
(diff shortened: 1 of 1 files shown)


DarkNami:
Sorry, I confirmd this is a ionCube bug. If source code encoded by ionCube PHP Encoder, a big int value will be cast as **float** not **int**.

No encoded file:
noencoded

(but this patch can working with encoded or not encoded file.)

Issue #197, opened 2018-12-21
Blade cannot work properly, does not support custom directives/conditions


MarekGogol:
Hello,

thank you for you awesome package. I have som updates/fixes to your blade compiler.

If your application does have custom directives/conditions, your compiler does not work properly.

version: *v4.6.1*

## Reproduction of error:

**AppServiceProvider.app**
In this provider we register our custom directive.
```php
<?php

namespace App\Providers;

use Illuminate\Support\ServiceProvider;
use Illuminate\Support\Facades\Blade;

class AppServiceProvider extends ServiceProvider
{
    /**
     * Register any application services.
     *
     * @return void
     */
    public function register()
    {
        Blade::directive('title', function ($text) {
            return "<?php echo $text ?>";
        });
    }
}
```

**layout.blade.php**
My blade template with one working title and one buggy title... (Our application has custom directives for seo purposes...)
```blade
<h1>{{ _('This is my article, this title works OK') }}</h1>
<h2>@title( _('This is my second title, and this does not work properly') )</h2>
```

## Solution:
Problem is in your blade extractor, where you are initializing new `Illuminate\View\Compilers\BladeCompiler` instance, instead of using existing facade `Illuminate\Support\Facades\Blade` which has registered all blade directives/conditions. When you initialize new BladeCompiler instance, then all custom directives binded into facade are missing. In this case your compiled template will not be properly compiled as in regular laravel response.

**Solution number 1.**
You can just simple use Facade for compiling blade template. Here is updated your blade compiler.
```php
<?php

namespace Gettext\Extractors;

use Gettext\Translations;
use Illuminate\Support\Facades\Blade as BladeFacade;

/**
 * Class to get gettext strings from blade.php files returning arrays.
 */
class Blade extends Extractor implements ExtractorInterface
{
    /**
     * {@inheritdoc}
     */
    public static function fromString($string, Translations $translations, array $options = [])
    {
        $string = BladeFacade::compileString($string);

        PhpCode::fromString($string, $translations, $options);
    }
}
```

**Solutions number 2**
But I am not sure, why you are using this blade compiler instead of using facade. Maybe better solution for some users who is using Blade without laravel will be this. ( Maybe blade without laravel is 💩  :)), so we do not need this solutions. )
```php
<?php

namespace Gettext\Extractors;

use Gettext\Translations;
use Illuminate\Filesystem\Filesystem;
use Illuminate\View\Compilers\BladeCompiler;

/**
[...]

PR #198, opened 2018-12-22
added option in blade extractor for using blade facade

MarekGogol:
Fixed missing custom directives in default Blade compiler instance. We added option to use Facade with all binded parameters for custom blade directives and conditions support.

Issue #197


MarekGogol:
@oscarotero Please, could you make release with tag?

oscarotero:
Tag released.

Issue #200, opened 2019-03-21
\n in JS files: backslash \ is stripped


jakubjo:
Thank you for this awesome package!

I've encountered following issue:

When scanning JavaScript files the backslash of new line sequences is stripped.

```js
gettext("First Line\nSecond Line\nThird Line");
```

Results in following msgid:

`msgid "First LinenSecond LinenThird Line"`

When I remove the case in JsFunctionsScanner:39 extracting is working as intended.
 https://github.com/oscarotero/Gettext/blob/93176b272d61fb58a9767be71c50d19149cb1e48/src/Utils/JsFunctionsScanner.php#L39

I'd submit a PR, but I'm unsure how to fix this properly and do not really understand all whats happening in JsFunctionsScanner.

oscarotero:
I guess we need to check if the current status is `simple-quote` or `double-quote` and make a break if so.
Something like this: https://github.com/oscarotero/Gettext/blob/93176b272d61fb58a9767be71c50d19149cb1e48/src/Utils/JsFunctionsScanner.php#L55

Thanks!

jakubjo:
We use \n also in single quotes in JS. I believe JS does not require double quotes around new line sequences. 

I've submitted a PR. In our case the issue is fixed and everything is working as intended. I'm curious what you'll say.

PR #201, opened 2019-03-21
Fix stripping backslash from new line characters in JS files (#200)

jakubjo:
This is the PR for previously created issue #200.

I'm not quite sure if its enough.
Everything went fine in my tests with 756 strings accross 89 files.

### DIFF
--- a/src/Utils/JsFunctionsScanner.php
+++ b/src/Utils/JsFunctionsScanner.php
@@ -104,7 +104,7 @@ class JsFunctionsScanner extends FunctionsScanner
      */
     private function getArguments(Node $node)
     {
-        return array_map(function($argument) {
+        return array_map(function ($argument) {
             return $this->getArgumentValue($argument);
         }, $node->getArguments());
     }
--- a/src/Utils/JsFunctionsScanner.php
+++ b/src/Utils/JsFunctionsScanner.php
@@ -37,11 +37,21 @@ class JsFunctionsScanner extends FunctionsScanner
 
             switch ($char) {
                 case '\\':
-                    $prev = $char;
-                    $char = $next;
-                    $pos++;
-                    $next = isset($this->code[$pos]) ? $this->code[$pos] : null;
-                    break;
+                    switch ($this->status()) {
+                        case 'simple-quote':
+                            break 2;
+                        case 'double-quote':
+                            break 2;
+
+                        default:
+
+                            $prev = $char;
+                            $char = $next;
+                            $pos++;
+                            $next = isset($this->code[$pos]) ? $this->code[$pos] : null;
+
+                            break;
+                    }
 
                 case "\n":
                     ++$line;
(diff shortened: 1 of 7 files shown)


src/Utils/JsFunctionsScanner.php
```diff
@@ -37,11 +37,18 @@ public function getFunctions(array $constants = [])
 
             switch ($char) {
                 case '\\':
-                    $prev = $char;
-                    $char = $next;
                     $pos++;
+
+                    if ($next !== 'n') {
```
oscarotero: The escaped character cannot be only `\n` but any character (like, for example, `\t`).
I think we should accept any character inside double or single quotes.

jakubjo: I'll look into it. So essentially we could just remove the whole `\\` case from the switch statement?

oscarotero: I guess just adding something like this:

```php
switch ($this->status()) {
    case 'simple-quote':
    case 'double-quote':
        break;
    default:
        //everything else
}
```

jakubjo:
Fixed it -- tests are running fine.

Screenshot 2019-03-25 00 29 53

jakubjo:
Unfortunately we've encountered additional issues where js files are not parser properly.
I'll investigate further and come back with more informations later.

jakubjo:
I've migrated the JS parser to use @mck89's peast parser.

The parser had issues with various, presumably legacy styled, files. Some strings would not be extracted. 
I've added a failing test first, so you're able to verify the bug by checking out commit bc85e89.

## Alle tests are now green. 
Additionally I've tested it on 285 different JS files ranging from legacy ugly pre ES2015 styled files to newest babel based styled files.

oscarotero:
Sorry but I wouldn't like to add more dependencies to this package, specially if it's something required only by one extractor (and most users wont use).

If you want to use `peast` (and looks a great project), I recomend to create a different package for this extractor and use Gettext as dependency.

jakubjo:
The elegance of your package lies in the all-in-one solution.

Why care about an additional dependency? Is there any significant issue with an additional dependency? That's the soul of open source: Don't reinvent the wheel. 

> and most users wont use

But essentially you don't know. The extractor is part of your package and has a bug in the current implementation. 
Which would be fixed with this PR.

**So final words were spoken?**

oscarotero:
> The elegance of your package lies in the all-in-one solution.

To me, the elegance of this package is that it can be extended by the community with other features. For example, there are twig translators, symfony bundlers, laravel and cake integrations, wordpress, cli tools, etc. I love that people want to create and maintain these packages, but I wouldn't like to have it integrated in this package. The more dependencies and features, the more work and time is needed to maintain everything up to date.

The extractors provided do not require dependencies, because many people install this package only for the po->mo conversion, or any other specific feature, so I do not want to force them to install a package that they do not need, or even worse, may cause conflicts with their dependencies.

I made a commit to solve this issue. Let me know if it works fine now.

jakubjo:
Thank you very much for your efforts. Its perfect! 👍

oscarotero:
Great! I close this.

PR #205, opened 2019-04-15
VueJS HTML tag support in extractor

gielwijgergangs:
This patch supports extracting VUE.JS gettext in HTML tags with:

- <translate> as tagname
- v-translate as attribute

It also supports context, comments and plural. 

See https://github.com/Polyconseil/vue-gettext how this is commonly used


oscarotero:
thanks!

PR #211, opened 2019-06-21
add source & target  language headers

ScottAustin:


oscarotero:
Hi!
Thanks, it looks good. Just two things:

- The header `X-Target-Language` is not neccesary because the value is stored in `Language`. 
- Don't you need to make any change in the Xliff extractor?

PR #215, opened 2019-07-12
Trim tab characters from comment lines

ocean90:
Related report: https://github.com/wp-cli/i18n-command/issues/169

When a PHP comment is indented with a tab character it currently doesn't get extracted. Adding the tab to the characters to strip will fix this.


swissspidy:
Looks like this breaks `Gettext\Tests\AssetsTest::testTwig
twig/Po` as the line numbers change. Haven't checked whether the assertions were actually wrong or if this is a regression.

ocean90:
@swissspidy Those are unrelated based on the build status of the master branch.

swissspidy:
Ah yes, indeed! 👍

ocean90:
https://github.com/oscarotero/Gettext/pull/216 updates the incorrect line numbers.

oscarotero:
Thanks 👍

PR #223, opened 2019-09-09
Multiple domain scan

briedis:
We use this library in a pretty big project which has hundreds (probably more than a thousand) files we need to scan. To scan the whole source we need 15+ minutes which is a lot, especially if the code changes very rapidly.

The main issue is that to scan each domain, the library has to parse a file for each domain separately. So my solution offers the option to scan a single file only once for all the domains in a single go.

I benchmarked my solution when you need to scan 20 domains against the old way where you scan each domain individually and the difference was 12x. And that was for a simple PHP file with <50 lines. The speed gain should be even larger for JS files which need complex parsing, a lot of string operations.

I tried to implement this as an additional feature so there would be no backward compatibility issues. Currently it supports PHP/JS(VueJS) parsing.

Some minor refactoring is included too.


briedis:
Specified trusty distribution in the Travis.yml file, because 5.4 and 5.5 was not building.
See: https://travis-ci.community/t/php-5-4-and-5-5-archives-missing/3723/4

oscarotero:
Thanks, it's really good idea.
Good job!

Issue #224, opened 2019-09-16
Duplicate translations with the same XLIFF unit ID are not maintained


asmecher:
This is a knock-on issue from https://github.com/oscarotero/Gettext/issues/220. (I hope you won't regret adding that already!)

An XLIFF file may contain several units with the same translation content... (silly example but you get the point)
```
<unit id="am.permanently">
  <segment>
    <source>am</source>
    <target>soy</target>
  </segment>/unit>
<unit id="am.temporarily">
  <segment>
    <source>am</source>
    <target>estoy</target>
  </segment>
</unit>
```
However, when loading the translation from the XLIFF file, only one gets loaded into `$translations`. This is because the two will have the [same ID](https://github.com/oscarotero/Gettext/blob/a545ffb99a597178132f15a91961c0df30a26535/src/Translation.php#L29), based only on the source text ("am"), without considering the unit ID.

Thus trying to save the loaded XLIFF again will drop one of the two entries.

asmecher:
@oscarotero, I am happy to code up a solution, but as with https://github.com/oscarotero/Gettext/issues/220 I'd like a little guidance about the best option because I'm not familiar with how the translation IDs are used elsewhere in the library.

oscarotero:
Mmm, I don't know how to solve this.

On the one hand, the unit id is not unique, on the other hand, the source is not unique either. I guess the uniqueness of a translation in Xliff is a sort of combination of unit id + source?

asmecher:
> I guess the uniqueness of a translation in Xliff is a sort of combination of unit id + source?

I don't think XLIFF has any requirements of uniqueness involving the source text. The [XLIFF 2.0 spec](http://docs.oasis-open.org/xliff/xliff-core/v2.0/os/xliff-core-v2.0-os.html) does talk about uniqueness of IDs, but that's pretty much it. For example, by my read, each unit (=`Translation` class instance) can be uniquely identified by a combination of unit ID and file ID (see section 3.1).

I think it'll be necessary to inform the mechanism for ID generation (currently the static `generateID` function) based on the format of the file being sourced.

oscarotero:
In gettext, we have the following structure:

- domain
- context
- original

Right now, the equivalent in Xliff would be:

- domain (the `id` of the file)
- context (the `<node category="context">` element)
- original (the `<source>` element)

A possible solution could be to get the `id` of the  unit as the msgid, instead the source, and store the source as a comment. So in a exported .po file, we could get:

```po
[...]

PR #225, opened 2019-09-19
#224 Allow for setting custom ID to avoid translation clobbering

asmecher:
Adds a `Translation::setId` function, with support for the Xliff extractor to call it with the Unit ID.

This permits the library to better conform to xliff's uniqueness requirements (the same source text can be translated several times with different unit IDs) while still supporting PO's uniqueness requirements (the same source text cannot appear twice in the same context).

Note that this means XLIFF files converted to PO files may lose some translations that appear several times with the same source text. (This is necessary to avoid a regression of https://github.com/oscarotero/Gettext/issues/48.)

Note that I did not modify e.g. the PO extractor to use `setId`, even though it supports round-tripping XLIFF unit IDs. These unit IDs can still be read out from the comments or using `\Gettext\Genreators\Xliff::getUnitID`. (This felt like the best design choice.)

Note the warnings in `Translation::setId` and `Translations::find`: Translations with custom IDs (e.g. XLIFF unit IDs) may not be found by the `Translations::find` function, as it searches using IDs generated by the `Translation::generateId` function.

@oscarotero, this may break `Translation::find` for  XLIFF files that were previously generated by this library, as they will use a `md5` sum for the unit ID.

### DIFF
--- a/src/Extractors/Xliff.php
+++ b/src/Extractors/Xliff.php
@@ -11,11 +11,18 @@ use SimpleXMLElement;
  */
 class Xliff extends Extractor implements ExtractorInterface
 {
+
+    public static $options = [
+        'unitid_as_id' => false
+    ];
+
     /**
      * {@inheritdoc}
      */
     public static function fromString($string, Translations $translations, array $options = [])
     {
+        $options += static::$options;
+
         $xml = new SimpleXMLElement($string, null, false);
 
         foreach ($xml->file as $file) {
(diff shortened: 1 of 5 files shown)


oscarotero:
> this may break Translation::find for XLIFF files that were previously generated by this library, as they will use a md5 sum for the unit ID.

Yes, I'd add the `$translation->setId($unitId);` optional. Something like:

```php
$translations = Translations::fromXliffFile('translations.xliff', ['unitid_as_id' => true]);
```

And by default, set to `false`.

asmecher:
@oscarotero, thanks, I've added that option and adjusted the tests accordingly. https://github.com/oscarotero/Gettext/pull/225/commits/3384000cbfcb9e4dd35ea039b441647423580d58

oscarotero:
Thanks!

asmecher:
Many thanks for your sensible stewardship, @oscarotero!

PR #227, opened 2019-10-01
 Sort output option added to Po generator 

vaites:
Separated pull request as requested ;)


src/Generators/Po.php
```diff
@@ -27,6 +29,12 @@ public static function toString(Translations $translations, array $options = [])
 
         $lines[] = '';
 
+        if ($options['sortOutput'] == true) {
+            $translations->uasort(function (Translation $translation1, Translation $translation2) {
+                return strcmp($translation1->getOriginal(), $translation2->getTranslation());
```
briedis: I don't understand - sorting by comparing original with the translation?

vaites: My fault. The comparison must be this: 
```php
return strcmp($translation1->getOriginal(), $translation2->getOriginal());
```
Sorry. Must I send a fix or can you change it?

Issue #231, opened 2019-10-17
Extensibility issues


schlessera:
We're using this library in `wp-cli/i18n-command` as the new basis to improve the translations build step for WordPress Core.

To extend it so that it does what we need, we need to jump through some hoops in various places of the code, because extensibility is made difficult by a few practices like ignoring late static binding or directly instantiating specific classes within a big chunk of logic.

Would you be open to a pull request that (in a generally backwards compatible way) improves extensibility across the board?

This would allow us to make the changes we need without copy-pasting big chunks of code and needing to keep them in sync.

The main changes I would like to introduce are:
- For non-final classes, default to `static` instead of `self` (late static binding) => this allows for all the code to properly support any extensions down the line.
- For instantiations of classes, extract the actual classes either into a method/constant that can be overridden, or a factory

oscarotero:
Yes, sure.
Just try to do not include so many changes in the same PR, so it's easier to review :)

schlessera:
I'll split it up into multiple PRs as needed then.

schlessera:
@oscarotero I have created 4 separate PRs with the changes I think are needed to make the classes properly extensible while still keeping everything pragmatic.

I have added lots of reasoning in commit messages to explain each of the changes, and split it up into separate commits so they can be cherry-picked if needed.

Looking forward to reading your feedback.

oscarotero:
@schlessera The changes are right, so I just merge them and will be included in the next release.

schlessera:
Awesome, thanks so much for the quick iteration!

oscarotero:
FYI, I'm working in a new (rewritten) v5 version to solve some of the issues of this library:

- Difficult to extend
- It contains many extractors/generators that not all people need
- Some spaguetti code

The idea is splitting this library in short packages that people can install if needed. The core package `gettext/gettext` only contains support for PO/MO files in addition to php arrays and code. The translator may be published as `gettext/translator`, same as other formats like `gettext/xliff`,  `gettext/twig`, etc.

If you want to take a look to provide feedback, you can do here: https://github.com/oscarotero/Gettext/tree/v5-dev

This is a work in progress yet.

rgpublic:
[...]

PR #234, opened 2019-10-17
Remove direct references to class `Translation`

schlessera:
To make the class `Translation` extensible without introducing bugs, the `Translations` class was now adapted to accept a class name of a Translation object to use.

This makes sure that when you extend `Translation`, you can enforce use of that extension even when `Translations` needs to create new instances of it.

`Translation` also has a `create()` method now so that when you have an instance of a translation you can easily create a new one without needing to know the exact class name.

Related issue: #231

### DIFF
--- a/src/Translations.php
+++ b/src/Translations.php
@@ -120,13 +120,19 @@ class Translations extends ArrayObject
         ],
     ];
 
-    private $headers;
+    protected $headers;
+
+    protected $translationClass;
 
     /**
      * @see ArrayObject::__construct()
      */
-    public function __construct($input = [], $flags = 0, $iterator_class = 'ArrayIterator')
-    {
+    public function __construct(
+        $input = [],
+        $flags = 0,
+        $iterator_class = 'ArrayIterator',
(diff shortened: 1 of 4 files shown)


src/Translations.php
```diff
@@ -473,4 +481,18 @@ public function mergeWith(Translations $translations, $options = Merge::DEFAULTS
...
+     *
+     * @param string $context  The context of the translation
+     * @param string $original The original string
+     * @param string $plural   The original plural string
+     * @return Translation New Translation instance
+     */
+    public function createNewTranslation($context, $original, $plural = '')
+    {
+        $factoryMethod = "{$this->translationClass}::create";
+        return $factoryMethod($context, $original, $plural);
```
swissspidy: That's PHP 7+ only I think?

schlessera: Yes, apparently. Will rewrite.

schlessera: Resolved via https://github.com/oscarotero/Gettext/pull/234/commits/b66e289796f1c7910f4b5b205db2cdfe09bb23ce

PR #238, opened 2019-11-14
Fix PHP error when scanning for single domain, but other strings found.

briedis:
Fixes this error: https://github.com/php-gettext/Gettext/issues/228

There was a wrong `if` condition. Created a test which fails. Patch fixes this simple case.


oscarotero:
Thanks!

PR #241, opened 2019-12-01
fixed diacritics for vuejs extractor from <template> element

MarekGogol:
Hello,

At the beginning, thanks for this awesome package. 

Version: v4.8.1
PHP: 7.2
Env: Mac OS - Cataline, Ubuntu server...

If you want import text with diacritics from vuejs `<template>` element with PHP extractor, output will be with buggy diacritics...

**VueJS**
```html
<template>
    {{ __('text with diacritics čťľďáé') }}
</template>
```

**PHP**
```php
Translations::fromVueJsFile(...)
```

Response will be **x04text with diacritics ÄÅ¥Ä¾ÄÃ¡Ã©** , but should be **text with diacritics čťľďáé**.

Fixed, and added tests... Hope everything is okay. Please could you publish new tag in v4?

Thanks.

### DIFF
--- a/src/Extractors/VueJs.php
+++ b/src/Extractors/VueJs.php
@@ -165,7 +165,7 @@ class VueJs extends Extractor implements ExtractorInterface, ExtractorMultiInter
         $dom = new DOMDocument;
 
         libxml_use_internal_errors(true);
-        $dom->loadHTML($html);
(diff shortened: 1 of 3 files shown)


briedis:
Just a note: Duplicate, but a slightly different solution: https://github.com/php-gettext/Gettext/pull/242

oscarotero:
Merged #242 that fixes this.

PR #242, opened 2019-12-02
Fix UTF-8 handling for VueJs scanner.

briedis:
DOMDocument does not handle UTF8 well without using htmlentities() or prepending XM encoding tag.

Added a fix and a test.

### DIFF
--- a/src/Extractors/VueJs.php
+++ b/src/Extractors/VueJs.php
@@ -165,7 +165,10 @@ class VueJs extends Extractor implements ExtractorInterface, ExtractorMultiInter
         $dom = new DOMDocument;
 
         libxml_use_internal_errors(true);
-        $dom->loadHTML($html);
+
+        // Prepend xml encoding so DOMDocument document handles UTF8 correctly.
+        // Assuming that vue template files will not have any xml encoding tags, because duplicate tags may be ignored.
+        $dom->loadHTML('<?xml encoding="utf-8"?>' . $html);
 
         libxml_clear_errors();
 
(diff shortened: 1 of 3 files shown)


MarekGogol:
luck?  #241 :)

briedis:
@MarekGogol goddamnit. We just noticed this issue on Friday.

What do you think if this solution (prepending XML encoding tag)?

Calling htmlentities to convert characters seems a bit risky, not sure I trust PHP enough with encoding conversions (bad experience). Ran some benchmarks - xml tag approach is a tiny fraction faster too.

MarekGogol:
I think your solution is good enough. VueJs template would not consists of <xml> tag, as you sad...

To be honest, I am not 100% sure which one is better. I just bring solution from link bellow. But your might be faster, so we can apply yours...
https://stackoverflow.com/questions/8218230/php-domdocument-loadhtml-not-encoding-utf-8-correctly?answertab=active#tab-top

We should wait for the author of package, and we will see what next.

briedis:
@MarekGogol Yeah, that is the same SO post I was looking at 😂 

I'm really happy that people are using the vue scanner and even contributing. Was not expecting that! (I was the author for vue scanner code)

briedis:
@oscarotero sorry to bother again, but without the fix package is unusable in production. 😬

MarekGogol:
> @oscarotero sorry to bother again, but without the fix package is unusable in production. 😬

So thanks @briedis ! VueJs is one of my favourite scanner :). I just didn't realize for one year that all my existing translations in client projects are crap :D. This week I was working on my project, and I find it out 😄 

image

oscarotero:
Ok. I have a package for that (https://github.com/oscarotero/html-parser/blob/master/src/Parser.php) but wouldn't like to add a dependency for this (maybe in v5, because we have a different package for each scanner).
I'm merging this.

PR #254, opened 2020-04-09
Update Scanner.php

vinodraut:
handled generated notice on php 7.4 if its not file by by insuring it should be file, which further generate fatal error as fread returns false.

### DIFF
--- a/src/Scanner/Scanner.php
+++ b/src/Scanner/Scanner.php
@@ -77,6 +77,12 @@ abstract class Scanner implements ScannerInterface
      */
     protected static function readFile(string $file): string
     {
+        // handled generated notice on php 7.4 if its not file
+        // by insuring it should be file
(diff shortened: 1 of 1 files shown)


vinodraut:
> Thanks for your contribution but I think that if the value is not a file, it should throw an Exception, as it does in line 89.

You are right, but I have weird experience on PHP 7.4.3+ where some items creates a problem with error 21 and further execution gets stop. But with same Glob to work on same code it works fine on below 7.4.3.

As this function is to read the files, I think we should do it as precautionary so there should not break the functionality.

Thank you

oscarotero:
What is the line that generates the code? `filesize` or `fopen` ?

If it's `filesize` maybe we can change this code to something more modern, without need to use filesize:

```php
$content = stream_get_contents($fd);
```

Issue #208, opened 2019-06-05
Way to generate string location path relative?


devnix:
I was wondering if there is any possibility of generating a relative path when I'm generating a .pot file when scanning files. Por example, not having lines like
```po
#: /var/www/html/httpdocs/index.php:539
```
and getting something more like
```po
#: ../index.php:539
```
if the pot file is being generated on /var/www/html/httpdocs/translations (just a fast example)

oscarotero:
This is not implemented currently.
To do it, we should edit this line: https://github.com/oscarotero/Gettext/blob/master/src/Extractors/Extractor.php#L17
For example, we could provide an option to define the base path used to calculate the relative path of all files, like this:

```php
$translations = Translations::fromPhpCode($file, ['basepath' => __DIR__]);
```

I don't think to have time to do this, if you want to help with a pull request, I'll happily review it.

devnix:
I would like to see if I can get some time for this. What would be the best default option? I personally would find the relative path the best option.

oscarotero:
For backward compatibility, the default option should be absolute path. Relative path will be generated only if `basepath` option is passed.

devnix:
That's what I was figuring. Let's see if I can make a PR this weekend.

So, we are limited to passing the base path to calculate a relative one, right? There is [this](https://github.com/symfony/filesystem/blob/56bf892e9e9138000e10f3148ffeb8fecdbb2d5c/Filesystem.php#L447) neat method from [symfony/filesystem](https://github.com/symfony/Filesystem) but I dont know If you are willing to include another dependency just for this method.

If we are not going to add that component as a dependency, we would be breaking any license if we just copy and paste that method?

devnix:
This could be a good usage?

```php
$translations->addFromJsCodeFile($file, [
    'base_path' => '/var/www/html'
]);
```

oscarotero:
> This could be a good usage?

Yes, perfect.

> ...I dont know If you are willing to include another dependency just for this method.

No, this should be implemented without include any dependency

> we would be breaking any license if we just copy and paste that method?

I think we can simplify this method, not only copy and paste as is. Anyway, the license is MIT so I think there's no problem, but I'd include a php comment referencing to the original method.

maicol07:
Any news on this one?

mjrider:
for ppl wanting this, this is the code we have running for a while now
```
$finder = new Symfony\Component\Finder\Finder();

$finder
    ->files()
[...]

Issue #143, opened 2017-03-24
[Q] How to turn off locations of strings in po file?


zaigham:
How can I achieve the following with `\Gettext\Generators\Po::toFile` ?

```
'--no-location’
Do not write ‘#: filename:line’ lines.
```
I want to remove strings locations from code files in generated PO file.
Thanks.

Ref: https://www.gnu.org/software/gettext/manual/html_node/msgattrib-Invocation.html

oscarotero:
Currently this is not supported, but it's easy to implement.
Meanwhile, you can remote the references before generate the PO file:

```php
foreach ($translations as $translation) {
    $translation->deleteReferences();
}

$translations->toPoFile('locales.po');
```

oscarotero:
I've added this option to po generator (ref: 97d7b7d16f9cc57e89ecdf2cd61dced0796197ca):

```php
$translations->toPoFile('locales.po', ['noLocation' => true]);
```

It will be released in the new version.

PR #256, opened 2020-04-11
Add configuration for location of strings in Scanner

Arzaroth:
This addresses both #208 and #143 

I didn't provide tests for this PR since there is no concrete implementation of CodeScanner in this repository

### DIFF
--- a/src/Scanner/CodeScanner.php
+++ b/src/Scanner/CodeScanner.php
@@ -14,6 +14,10 @@ abstract class CodeScanner extends Scanner
 {
     protected $ignoreInvalidFunctions = false;
 
+    protected $enableLocations = true;
+
+    protected $relativeBasePath = null;
+
     protected $commentsPrefixes = [];
 
     protected $functions = [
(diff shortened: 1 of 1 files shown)


oscarotero:
Hi, thanks for your work with these issues!

I'd rather have these two issues in two different pull requests, because I have some doubts:

- I'm agree with adding the option to enable or disable the translations locations, but I don't like the option name because it's confusing. I'd rename to `addReferences()`, to keep consistency with the `$translations->getReferences()` function.
- I'm not sure about the relative path, and this is why I'd like to have a different pull request. The reason is that perhaps it add too much responsabilities to a class that is created to scan an unique file. If anyone want to use a different file path, the solution is simple: passing directly the content and filename. For example:

```php
$contents = file_get_contents($file);
$relativePath = get_the_relative_path($file);

$translations = $scanner->scanString($contents, $relativePath);
```

This function is already present in the interface (https://github.com/php-gettext/Gettext/blob/master/src/Scanner/ScannerInterface.php#L21) and this way is much more flexible, so I'd wouldn't add this feature (at least, for now).

Arzaroth:
Points taken.

I'll close this PR and submit two new ones.

Regarding relative paths, I didn't realize what you suggested was possible. It is indeed a nice workaround and I understand your argument on this matter.

PR #259, opened 2020-04-24
Relative strings location path

Arzaroth:
Following the discussion in #256.

This addresses #208.


oscarotero:
As commented in https://github.com/php-gettext/Gettext/pull/256#issuecomment-618588669 I think this change introduces unnecessary complexity to the Code Scanner, because the function `scanString()` allows to achieve this.

In fact, `scanFile()` uses `scanString()` internally: https://github.com/php-gettext/Gettext/blob/master/src/Scanner/Scanner.php#L41

So I'm going to pause this for now.
Thanks, anyway!

PR #263, opened 2020-11-17
Multiline disabled translations

lbfeenix:
If .po file contain multiline disbled translations like this:

#~ msgid "Last agent hours-description"
#~ msgstr ""
#~ "How many hours in the past can system look at finding the last agent? "
#~ "This parameter is only used if 'Call Last Agent' is set to 'YES'."

only the first line was taken, this fix allow mutiline disabled translations.

### DIFF
--- a/src/Loader/PoLoader.php
+++ b/src/Loader/PoLoader.php
@@ -30,8 +30,7 @@ final class PoLoader extends Loader
             ) {
                 if (substr(trim($nextLine), 0, 1) === '"') { // Normal multiline
                     $line = substr($line, 0, -1).substr(trim($nextLine), 1);
-                }
-                if (substr(trim($nextLine), 0, 4) === '#~ "') { // Disabled multiline
+                } elseif (substr(trim($nextLine), 0, 4) === '#~ "') { // Disabled multiline
                     $line = substr($line, 0, -1).substr(trim($nextLine), 4);
                 }
                 $nextLine = next($lines);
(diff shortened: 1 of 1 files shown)


src/Loader/PoLoader.php
```diff
@@ -26,9 +26,14 @@ public function loadString(string $string, Translations $translations = null): T
...
             while (substr($line, -1, 1) === '"'
                 && $nextLine !== false
-                && substr(trim($nextLine), 0, 1) === '"'
+                && (substr(trim($nextLine), 0, 1) === '"' || substr(trim($nextLine), 0, 4) === '#~ "')
             ) {
-                $line = substr($line, 0, -1).substr(trim($nextLine), 1);
+                if (substr(trim($nextLine), 0, 1) === '"') { // Normal multiline
+                    $line = substr($line, 0, -1).substr(trim($nextLine), 1);
+                }
+                if (substr(trim($nextLine), 0, 4) === '#~ "') { // Disabled multiline
```
oscarotero, line 34: Here I think it's better a `else if`, so if the first condition is true, this second is not evaluated

oscarotero (changes requested):

oscarotero:
Thanks @lbfeenix 
I've added this test case: https://github.com/php-gettext/Gettext/commit/219a7e61b074ea2e7dc1ee3341958f4c81785a77, so please make sure it is passed.
I've also leave one comment in your code.

oscarotero (approved):

PR #266, opened 2021-03-10
PHP 8 compatibility (Gettext 4.x branch)

MichaelHoste:
It just removes one char for PHP 8 compatibility (at least for [our specs](https://github.com/translation/laravel/tree/master/tests)).

The error fixed is `Argument #1 ($carry) must be passed by reference, value given`

-------

**Note**

We still use the Gettext 4.x branch as dependency in [translation/laravel](https://github.com/translation/laravel) (for [https://translation.io](https://translation.io/laravel)) to be able to keep using the Blade Extractor.

Do you know if someone already made a Blade Scanner compatible with 5.x?

Just for information, how hard do you think it would  be to create this scanner by using the code of your 4.x extractor?

Thanks for this great library!

### DIFF
--- a/src/Utils/FunctionsScanner.php
+++ b/src/Utils/FunctionsScanner.php
@@ -30,7 +30,7 @@ abstract class FunctionsScanner
         $translations = is_array($translations) ? $translations : [$translations];
 
         /** @var Translations[] $translationByDomain [domain => translations, ..] */
-        $translationByDomain = array_reduce($translations, function (&$carry, Translations $translations) {
+        $translationByDomain = array_reduce($translations, function ($carry, Translations $translations) {
             $carry[$translations->getDomain()] = $translations;
             return $carry;
         }, []);


oscarotero:
Hi, thanks for fixing this!

About Blade Scanner for Gettext 5.x, it should be very easy to replicate what we have in 4.x because it simply compile the template (here: https://github.com/php-gettext/Gettext/blob/4.x/src/Extractors/Blade.php#L29) and then run the PHP code scanner (here: https://github.com/php-gettext/Gettext/blob/4.x/src/Extractors/Blade.php#L32)

In 5.x, I imagine something like the following (not tested):

```php
class BladeScanner extends PhpScanner
{
    public function scanString(string $string, string $filename): void
    {
        // create a compiler somehow
        $compiler = new BladeCompiler();

        // convert Blade code to regular PHP code
        $code = $compiler->compileString($string);

        return parent::scanString($code, filename);
    }
}
```

We should provide a way to customize the `BladeCompiler` (as we have in 4.x with `$options['facade']` and other options like a path for cache, etc. But basically, it's just that.

If you want to work on this, it would be great, i can create the BladeScanner repository and add you as a contributor.

MichaelHoste:
Thank you for the prompt new release @oscarotero !

And thank you for the extra information about the Blade scanner.

I may give it a try in the future to be able to bump the dependency, but I guess I may have to change a lot of other small things in our code like `$translations::fromPoFile` or `$translation->toPoFile()`. That's not overly complicated but I don't have the time right now.

> If you want to work on this, it would be great, i can create the BladeScanner repository and add you as a contributor.

Maybe not today but when I'll have some code to show, that would be great, thanks! :-)

PR #271, opened 2021-07-12
v4: Prevent adding the same translator comment to multiple functions

swissspidy:
It's been a while since my last my PR to this project.

You might recall that we're using this library in WP-CLI for string extraction in all of WordPress core and WordPress plugins.

Unfortunately due to different minimum PHP version requirements (still supporting PHP 5.6+) we can't use the latest version of the library yet.

That's why this bugfix is target at the 4.x branch and I hope you consider merging it.

The bug that we're facing with v4 is that translator comments get added to multiple functions if they are all on the same line, instead of just to 1 function.

The tests should demonstrate this, but basically with PHP code like this:

```php
printf( /* translators: Hello */ __( 'World' ) );
__( 'No comment' );

printf( /* translators: Only used once */ __( 'Within printf: %s' ), __( 'placeholder' ) ); __( 'After printf' );
```

The two translator comments should only be added to `__( 'World' )` and `__( 'Within printf: %s' )`, respectively.


oscarotero:
Great. Thank you! 👍

Issue #282, opened 2022-07-06
Handling of defective .po files


jonasraoni:
If I try to load a `.po` file with the following content:

```po
msgid "message.id"
msgstr "this line is ok"
this line is missing the initial quote"
```

It works, but the defective line is ignored. It would be great if the package could log a warning or fail with an exception (this could be probably configurable with a "strict mode", disabled by default to keep the compatibility).

I'll take a look in the code, perhaps I can contribute with another PR if you think the idea is ok.

oscarotero:
Yeah, it makes sense to me. PR is appreciated, thanks!

jonasraoni:
Great, I'll take a look at the code and `.po` format, which is short and try to come up with something generic.
My objective is to also be able to use the package to validate/identify broken files.

jonasraoni:
FYI, I've checked the `.po` format and I think it's too superficial, it's missing a lot of small details 🤔
I'll address the things that I saw and come up with a proposal.

jonasraoni:
- #283

PR #283, opened 2022-07-10
#282 Add strict po loader

jonasraoni:
Hi @oscarotero!

I did some initial updates to cover my needs.

- I saw it was missing support for the "previous translation comment", so I've added the needed parts to the translator, generator and the `StrictPoLoader`.
- About the strictness, I've tried to follow the rules of `msgfmt -c` (e.g. line break/whitespace is optional almost everywhere) 🫠
- I was going to enhance the `PoLoader`, but it became too polluted with my checks and I decided to write a new parser...
- I barely changed code to handle the comments and the headers.
- I've tried to avoid using regexps, as the performance drops severely with large texts


jonasraoni:
I've tested some weird combinations, but forgot to move them to the unit test 🤦

src/Loader/StrictPoLoader.php
```diff
@@ -0,0 +1,349 @@
...
+
+    private function readNumber(): string
+    {
+        for ($data = ''; ctype_digit($this->getChar() ?? ''); $data .= $this->nextChar());
+        return $data;
+    }
+
+    private function readCommentString(): string
+    {
+        for ($data = ''; ($this->getChar() ?? "\n") !== "\n"; $data .= $this->nextChar());
```
jonasraoni, line 109: Here I think it's interesting to support different line breaks, so I could add the `\r`.

jonasraoni, line 109: Here I think it might be interesting to add a `\r` to support different line breaks (a user using `\r\n` will probably end up with the `\r` being added to the `$data`).

oscarotero:
Hi.
Thanks for this work!!
I'm not familiarized with the strict mode. What's the difference between strict mode and regular po? This will help me to understand the need to create a new different Po loader.

jonasraoni:
Hi!

There's no "strict" syntax, I just mimicked the behavior of the GNU tools, which I think is a good guide (given the [documentation](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) is missing details) :)

For example, the `msgfmt` works fine with this syntax (the new lines are required just to break comments):
```po
msgctxt"context"msgid"message"msgstr"translation"
```

But it fails if you try to parse things in the wrong order, with an invalid `msgstr` index (out of order/out of bounds), bad escaping, etc.

So in theory this could replace the current `PoLoader` class, but it will introduce surprises (exceptions) for your users.

So I'll wait for your feedback. In order to decrease the pain, I could just add a flag, to swallow the exceptions by default, and continue to the next translation... But the input vs output of a defective file won't be the same as the current class.

src/Loader/StrictPoLoader.php
```diff
@@ -0,0 +1,349 @@
...
+
+    public function loadString(string $string, Translations $translations = null): Translations
+    {
+        $this->data = $string;
+        $this->position = 0;
+        $this->translations = parent::loadString($string, $translations);
+        for ($this->newEntry(); $this->getChar() !== null; $this->newEntry()) {
+            while ($this->readComment());
+            $this->readContext();
+            $this->readId();
```
jonasraoni, line 31: FYI here there's one different behavior from the GNU's `msgfmt` related to simple comments (`#`, `#.`, `#,`, `#:`) dangling in the end of the file, `msgfmt` ignores them, while this parser will raise an exception.

oscarotero: Looking at this code, it seems like the order of the elements matters. Every translation need the comments first, then the context, id, plurals and translation in this specific order. Is that true?

jonasraoni: I remembered about my last observation :)
If you use the `#,` with unknown flags (e.g. `#,test`), the comment is dropped... It only keeps the known ones, such as `java-printf-format`

I just didn't investigate the behavior of the `#:`, so I'm not sure if it can be wiped off/generate errors when invalid.

oscarotero: Ok. I prefer PoLoader to be more permisive than `msgfmt`. I mean, accept code that would fail in `msgfmt` (or other tools) and output code that is compatible with these tools, like a kind of "fixer".
I think that the order of the elements shouldn't be important.
I don't know if its easy to make this new StrictPoLoader more lax. If not, I wouldn't replace the existing PoLoader (althought both loaders can be included).

oscarotero: Yes, I'm agree.
I'm fine including both.
Looking forward the tests before merge it.

jonasraoni: Ok! I'll write some tests :)

jonasraoni: Ok, I fixed the broken tests and added some for the other loader.

jonasraoni: Please, give me one more run 😅

jonasraoni: The tests passed (both loaders are sharing the tests that you've wrote)...

Tomorrow I'll take a look at the performance, then feel free to merge or ask fixes.

Given the amount of checks that I've added, I guess it will be difficult to make a miracle, but there's room for improvements...

oscarotero: Nice, thanks!

Let me know when you are confortable with the code, so I can review and merge :)

jonasraoni:
Ok @oscarotero, I'm done with the performance fixes.
In a small dataset with 22 files (564KB) the time went from 4s to .47s, while the `PoLoader` is taking .25s.

oscarotero:
Thanks @jonasraoni 

> If you think it's helpful, I could rebase the commits to mark the tests/BasePoLoaderTestCase.php as a file rename from tests/PoLoaderTest.php, just to improve the code history.

Yeah, it makes sense. But I think it's not so important, so whatever you want :)

One think that I would like is update the README.md file to include this new loader and explain the differences between both PO loaders (including also the difference in performance). As this info is more in your head, could you please add this latest change before merging?

Thanks so much!

jonasraoni:
@oscarotero I've updated the README and the CHANGELOG 👌

oscarotero:
Thank you! 👍

Issue #285, opened 2022-11-20
add a ?? to string reader for php8.1 in /src/Utils/StringReader.php


lachlan-00:
https://github.com/php-gettext/Gettext/blob/3f7bc5ef23302a9059e64934f3d59e454516bec0/src/Utils/StringReader.php#L31

could we get this substr call edited to stop the depreciation errors for php8.1?

```
public function read($bytes)
    {
        $data = substr($this->str, $this->pos ?? 0, $bytes);

        $this->seekto($this->pos + $bytes);

        return $data;
    }

```

Or maybe just set it on construct?
```
    public function __construct($str)
    {
        $this->str = $str;
        $this->strlen = strlen($this->str);
        $this->pos = 0;
    }

```

The errors for translations on php8.1 are insanely loud when translating on php8.1
```
[Error] substr(): Passing null to parameter #2 ($offset) of type int is deprecated in file /var/www/music/vendor/gettext/gettext/src/Utils/StringReader.php(31)
```

I will be looking at moving to gettext5 but i don't really understand the code enough yet to do this for ampache

oscarotero:
Can you create a PR with this fix? Thanks!

lachlan-00:
merged with #287

PR #287, opened 2022-11-20
Update StringReader.php

lachlan-00:
* Set a default value for pos
* closes issue #285

I thought it would be better to make the pos parameter have a default value rather than fall back to 0 on an error like i originally did.

added manually and tested on my home server.


oscarotero:
Perfect, thanks!

PR #299, opened 2024-07-24
Support for PHP >= 8.2

### LINKED ISSUE #298: PHP 8.3 Support
Dear Developers,

i love this tool and i use it in the most of my PHP projects.

**Could you tell me when PHP 8 will be supported?**

Thanx a lot :-)

pedro-mendonca:
Since PHP 8.2 that ksort() deals with int/string keys differently.
https://php.watch/versions/8.2/ksort-SORT_REGULAR-order-changes

Up to PHP 8.1 the int keys are sorted last.
From PHP 8.2 the int keys are sorted first.

The current .mo test file has the original string `12345` in the end.
This was ok until PHP 8.2.

This PR makes sure that the strings of the generated `.mo` file are sorted by the original, comparing all keys as strings and outputs the strings with numbers before the ones with letters.

Another string is added to the test to make sure that the original `12345` is sorted before `15` (compared as strings, as it's supposed)

The result of the translations array before output to `.mo` file is:
- `12345`
- `15`
- `Original`
- `Other comment`

<hr>

Fixes #298


oscarotero:
It's perfect. Thank so much!

pedro-mendonca:
My pleasure 😃 
Thanks for this awesome project!

PR #300, opened 2024-10-08
Fix PHP 8.4 explicit nullable compatibility

cracksalad:
The upcoming release of PHP 8.4 will deprecate the use of implicit nullable types (`string $str = null`) and requires the explicit nullable type instead (`?string $str = null`). This PR modifies all occurences of the implicit nullable type such that this project should be compatible with PHP 8.4 at least in that regard. The modifications do not change any logic, just adding syntactic sugar, which will be required starting with 8.4.


oscarotero:
Thank you!

PR #304, opened 2025-04-19
Add PHPStan level 5

devnix:
Hi @oscarotero, I propose to add PHPStan slowly to improve the type coverage.

It may trigger some architectural questions, for example we are getting this error at level 0:

```
 ------ ----------------------------------------------------------------------------------- 
  Line   src/Flags.php                                                                      
 ------ ----------------------------------------------------------------------------------- 
  21     Unsafe usage of new static().                                                      
         🪪  new.static                                                                      
         💡 See: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static   
 ------ ----------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------- 
  Line   src/Headers.php                                                                    
 ------ ----------------------------------------------------------------------------------- 
  26     Unsafe usage of new static().                                                      
         🪪  new.static                                                                      
         💡 See: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static   
 ------ ----------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------- 
  Line   src/References.php                                                                 
 ------ ----------------------------------------------------------------------------------- 
  21     Unsafe usage of new static().                                                      
         🪪  new.static                                                                      
         💡 See: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static   
 ------ ----------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------- 
  Line   src/Translation.php                                                                
 ------ ----------------------------------------------------------------------------------- 
  30     Unsafe usage of new static().                                                      
         🪪  new.static                                                                      
         💡 See: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static   
 ------ ----------------------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------------------- 
  Line   src/Translations.php                                                               
 ------ ----------------------------------------------------------------------------------- 
  25     Unsafe usage of new static().                                                      
         🪪  new.static                                                                      
         💡 See: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static   
 ------ -----------------------------------------------------------------------------------
 ```
 
 It's not a big deal to fix, but making the class or the constructor final would be a breaking change, and I have no idea if it follows how you want to keep evolving this library.

### DIFF
--- a/src/Loader/MoLoader.php
+++ b/src/Loader/MoLoader.php
@@ -125,9 +125,7 @@ final class MoLoader extends Loader
 
     private function readInt(string $byteOrder): int
     {
-        if (($read = $this->read(4)) === false) {
-            return 0;
-        }
+        $read = $this->read(4);
 
         $read = (array) unpack($byteOrder, $read);
 
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "brick/varexporter": "^0.3.5",
         "friendsofphp/php-cs-fixer": "^3.2",
         "oscarotero/php-cs-fixer-config": "^2.0",
-        "phpstan/phpstan": "^2.1"
+        "phpstan/phpstan": "^1|^2"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
         "squizlabs/php_codesniffer": "^3.0",
         "brick/varexporter": "^0.3.5",
         "friendsofphp/php-cs-fixer": "^3.2",
-        "oscarotero/php-cs-fixer-config": "^2.0"
+        "oscarotero/php-cs-fixer-config": "^2.0",
+        "phpstan/phpstan": "^2.1"
     },
     "autoload": {
         "psr-4": {
@@ -41,7 +42,8 @@
     "scripts": {
         "test": [
             "phpunit",
-            "phpcs"
+            "phpcs",
+            "phpstan"
         ],
         "cs-fix": "php-cs-fixer fix"
     }
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -1,0 +1,5 @@
+parameters:
+    level: 0
+    paths:
+    - src
+    - tests
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -1,5 +1,5 @@
 parameters:
-    level: 4
+    level: 5
     paths:
     - src
     - tests
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -1,5 +1,5 @@
 parameters:
-    level: 1
+    level: 4
     paths:
     - src
     - tests
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -1,5 +1,5 @@
 parameters:
-    level: 0
+    level: 1
     paths:
     - src
     - tests
--- a/src/Comments.php
+++ b/src/Comments.php
@@ -11,6 +11,8 @@ use ReturnTypeWillChange;
 
 /**
  * Class to manage the comments of a translation.
+ *
+ * @phpstan-consistent-constructor
  */
 class Comments implements JsonSerializable, Countable, IteratorAggregate
 {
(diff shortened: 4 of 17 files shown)

### CODE REVIEW

**Comment by devnix** [comment]
Of course, [`@phpstan-consistent-constructor`](https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static#add-%40phpstan-consistent-constructor-to-the-class) could be used, but making those classes final doesn't sound stupid to me

**Comment by oscarotero** [comment]
Hi @devnix 
Thanks for this! I really appreciate your work.

I don't want to introduce breaking changes so `@phpstan-consistent-constructor` sounds good to me.
Let me know when it's ready to merge and I'll review it.

**Comment by devnix** [comment]
Thank you! I'm goint to iterate a little bit and I'll ask you a couple of specific questions here and there around the code 😄

**Comment by devnix** [comment]
@oscarotero, is the `src/Scanner/FunctionsHandlersTrait.php` still useful? PHPStan detects it's not used in any class or tested at all, so I'd like to know if it's an untested utility for the library consumer or a leftover.

**Comment by devnix** [comment]
Also, I'm wondering if you would be OK with adding Rector in another pull request so it can help reach PHPStan level 6 faster, applying each ruleset in granular pull requests.

**Comment by oscarotero** [comment]
> is the `src/Scanner/FunctionsHandlersTrait.php` still useful? PHPStan detects it's not used in any class or tested at all, so I'd like to know if it's an untested utility for the library consumer or a leftover.

It's used by other packages, like [PHP-Scanner](https://github.com/php-gettext/PHP-Scanner/blob/master/src/PhpScanner.php#L13).

**Comment by oscarotero** [comment]
> Also, I'm wondering if you would be OK with adding Rector in another pull request so it can help reach PHPStan level 6 faster, applying each ruleset in granular pull requests.

I'm not familiarized with Rector but if it helps to improve the code quality, it's okay, as long as it keeps compatibility with PHP 7.2 and newer

**Comment by devnix** [comment]
>I'm not familiarized with Rector but if it helps to improve the code quality, it's okay, as long as it keeps compatibility with PHP 7.2 and newer

Excellent!

The pull request should be ready to pass the workflows and be reviewed. Thank you!

**Comment by devnix** [comment]
It looks like PHPStan will not be installed in versions lower than 7.4. It should not pose a problem since it's just a devDependency.

I guess we could invoke PHPStan from Docker and remove the dependency

**Comment by oscarotero** [comment]
Can we use 1.x for these versions? I mean `^1 | ^2`.

**Comment by devnix** [comment]
I think there is a conflict between the versions shipped into my machine of php-cs-fixer and phpcs, as there is no `composer.lock` for the project currently:

image

--- 

image

Also, I'm aware that it's well-known that running php-cs-fixer with different PHP versions can bring very different behaviors, so I might have a couple of suggestions that I can address here or in another pull request if you want:

- Persisting a `composer.lock` in the repository to have predictable versions of the development packages
- Run style checkers together, as well as running both fixers in the same fix command too
- Run style checkers only on the lowest supported version maybe?

**Comment by devnix** [comment]
Workflows should be green now: https://github.com/devnix/Gettext/actions/runs/14604976659/job/40971880629?pr=1

File: src/Loader/MoLoader.php [hunk 1]
```diff
@@ -125,9 +125,7 @@ private function seekTo(int $position): void
 
     private function readInt(string $byteOrder): int
     {
-        if (($read = $this->read(4)) === false) {
-            return 0;
-        }
+        $read = $this->read(4);
```
oscarotero, line 128: I'm not sure if this is equivalent.
If the `$read` variable is `false`, it returns `0`. The new code doesn't check this condition.

oscarotero (changes requested):

src/Loader/MoLoader.php
devnix: That's the neat part, the signature of that function is `private function read(int $bytes): string`, and PHPStan reports that if as `MoLoader::read() does never return false`

oscarotero:
Great work! Thanks you!

PR #306, opened 2025-05-02
Rector coverage level 2

devnix:


src/Scanner/CodeScanner.php
```diff
@@ -110,7 +110,7 @@ protected function addComments(ParsedFunction $function, ?Translation $translati
 
     protected function addFlags(ParsedFunction $function, ?Translation $translation): ?Translation
     {
-        if (empty($translation)) {
+        if (!$translation instanceof \Gettext\Translation) {
```
oscarotero, line 113: I don't understand this change. What's the benefit?

oscarotero: > Probably `if ($value === null)` would be a better option

Yes, I think is clearer.

oscarotero (approved):

oscarotero:
Perfect. Thank you!

PR #307, opened 2025-05-02
Phpstan level 6

devnix:


src/Comments.php
devnix, line 1: Hi @oscarotero, could you please confirm that these changes are correct?

devnix: Yeah, there are a lot of pending changes, this PR should type every attribute and define the types of arrays and generics. Right now I had this doubt

oscarotero: As long as it doesn't introduce breaking changes and it's still supported by PHP 7.2 and newer, I'm okay with this.

src/Comments.php
```diff
@@ -51,9 +61,7 @@ public function delete(string ...$comments): self
         foreach ($comments as $comment) {
             $key = array_search($comment, $this->comments);
 
-            if (is_int($key)) {
```
devnix, line 54: My [previous question](https://github.com/php-gettext/Gettext/pull/307/files#r2071832898) was more related to this change.

As far as I understand, should `Comments::$comments` be an array of integer keys only? If that is correct, this condition would be unnecessary :)

oscarotero: `$key` can be false, if the comment wasn't found.

